PR #17, opened 2016-12-06
Add packet loss indicator in scoreboard

execut4ble:
lol 3 commits because im bög and skrew up kode blät))


YaLTeR:
Is this actually useful?

execut4ble:
Yes, useful if you want to see how much loss noi has to make his whining justified.

Other than that, vanilla AG has it, so why not.

YaLTeR:
AG has an IRC client, so why not add that as well. I'm cleaning stuff up in process.

execut4ble:
Still useful to see if you or your opponent is lagging.

PR #18, opened 2016-12-07
Add packet-loss indicators in the scoreboard

execut4ble:
This time with less commits 👍


YaLTeR:
Except I would squash them into one commit anyway.

PR #38, opened 2017-04-26
Add cl_scores_alpha

execut4ble:
rofl)


cl_dll/hud_scores.cpp
```diff
@@ -46,6 +47,7 @@ int CHudScores::Draw(float time)
 		return 0;
 
 	const size_t rows_to_draw = static_cast<size_t>(cl_scores->value);
+	const int alpha = static_cast<int>(cl_scores_alpha->value);
```
YaLTeR: Make it a `uint8_t`, this way it automatically gets bounded to the allowed 0-255 interval. Otherwise, LGTM.

Issue #29, opened 2017-01-14
Weapon configs


execut4ble:
Add a way to have different configurations based on your currently equipped weapon i.e. when you have gauss, your crosshair would be a red dot but when you pull out your 9mmAR you get a green crosshair and your weapon would be in a left-handed spot.

Could do this with config files or some in-game preset system:
`cl_weaponcfg_gauss gauss.cfg` would use settings from gauss.cfg when you have gauss in your hands.

There might be a better way to do this, if config files are slow or bad in some way.

PR #39, opened 2017-05-02
Add CVARs to draw individual crosshair lines

dxtzero:
add a cvar for each of the crosshair lines so that they can be individually drawn or not.


YaLTeR:
What's the point?

dxtzero:
i dont like vertical lines in my crosshairs, maybe someone doesnt like horizontal lines, maybe they want the left and bottom lines only, or some other combination?

#29 could benefit from this.

is there a reason not to have this? (im not sure how to convey a tone of being inquisitive and not antagonistic with that question)

YaLTeR:
My only concern is that it adds complexity, but I guess it's fine.

dxtzero:
i hope not too much :)

thanks!

PR #56, opened 2018-05-13
Add circle thickness and outline

execut4ble:
Smoothness on high thickness values needs to be fixed, but I am not sure how to do that. Perhaps by implementing a max value limit relative to the circle radius?

### DIFF
--- a/cl_dll/hud_crosshairs.cpp
+++ b/cl_dll/hud_crosshairs.cpp
@@ -7,20 +7,21 @@ int CHudCrosshairs::Init()
 {
 	m_iFlags = HUD_ACTIVE;
 
-	cl_cross =               CVAR_CREATE("cl_cross", "0", FCVAR_ARCHIVE);
-	cl_cross_color =         CVAR_CREATE("cl_cross_color", "0 255 0", FCVAR_ARCHIVE);
-	cl_cross_alpha =         CVAR_CREATE("cl_cross_alpha", "200", FCVAR_ARCHIVE);
-	cl_cross_thickness =     CVAR_CREATE("cl_cross_thickness", "2", FCVAR_ARCHIVE);
-	cl_cross_size =          CVAR_CREATE("cl_cross_size", "10", FCVAR_ARCHIVE);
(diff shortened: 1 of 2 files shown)


cl_dll/hud_crosshairs.cpp
```diff
@@ -108,6 +109,19 @@ int CHudCrosshairs::Draw(float time)
 			gl.line(Vector2D(center.x - offset.x, center.y - offset.y + half_width), Vector2D(center.x - offset.x, center.y + offset.y - half_width));
 			gl.line(Vector2D(center.x - offset.x - half_width, center.y + offset.y), Vector2D(center.x + offset.x + half_width, center.y + offset.y));
 		}
+
+		// Circle
+		if (cl_cross_circle_radius->value && cl_cross_circle_thickness->value > 0.0f) {
```
YaLTeR: Why did you put this here and not in the circle drawing if below? This block does the exact same thing as the one below except with `cl_cross_outline->value` added to thickness (why?).

execut4ble: I wanted to separate the circle's outline like the other outlines are. It just draws a thicker (with outline value added to normal thickness) circle that works as an outline I guess. Would there be a better way of doing this?

YaLTeR: Oh, I see.

YaLTeR:
Hmm, so one way of doing this (and also solving the smoothness issue) is to add a function similar to `HudGL::compute_circle()` that returns the trapezoid points, which means that (when outline is enabled) instead of one point (`radius` multiplied by `cos(angle)` and `sin(angle)`) it would return four points:
- `(radius - half_thickness - outline_width)`
- `(radius - half_thickness)`
- `(radius + half_thickness)`
- `(radius + half_thickness + outline_width)`

Then add a function to draw a trapezoid (essentially the same as `HudGL::rectangle()` but takes in 4 points and just feeds them to those 4 `glVertex2f()` calls) and a function to draw the circle from trapezoid segments, which will take those computed trapezoids and draw them all in a loop.

YaLTeR:
Closing. Feel free to reopen if you pick this up or someone else feel free to open a new PR with the proper way.

PR #60, opened 2019-06-09
Added autorecord stop on match end

execut4ble:
cl_autorecord_stop 1 will stop demo recording 3 seconds into intermission (map end screen)

### DIFF
--- a/cl_dll/hud.h
+++ b/cl_dll/hud.h
@@ -586,11 +586,11 @@ public:
 	cvar_t  *m_pCvarStealMouse;
 	cvar_t	*m_pCvarDraw;
 	cvar_t	*m_pCvarDrawDeathNoticesAlways;
+	cvar_t	*m_pCvarAutostop;
 
 	int m_iFontHeight;
 
 	cvar_t* m_pCvarColor;
-	cvar_t* cl_autostop;
 	unsigned long m_iDefaultHUDColor;
 
 	void UpdateDefaultHUDColor();
--- a/cl_dll/hud.h
+++ b/cl_dll/hud.h
@@ -590,6 +590,7 @@ public:
 	int m_iFontHeight;
 
 	cvar_t* m_pCvarColor;
+	cvar_t* cl_autostop;
 	unsigned long m_iDefaultHUDColor;
 
 	void UpdateDefaultHUDColor();
--- a/cl_dll/hud_countdown.cpp
+++ b/cl_dll/hud_countdown.cpp
@@ -19,7 +19,7 @@ int CHudCountdown::Init()
 	name2[ARRAYSIZE(name2) - 1] = '\0';
 
 	cl_autorecord = CVAR_CREATE("cl_autorecord", "0", FCVAR_ARCHIVE);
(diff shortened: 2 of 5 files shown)


cl_dll/hud_redraw.cpp
```diff
@@ -127,6 +128,9 @@ int CHud :: Redraw( float flTime, int intermission )
 			// Take a screenshot if the client's got the cvar set
 			if ( CVAR_GET_FLOAT( "hud_takesshots" ) != 0 )
 				m_flShotTime = flTime + 1.0;	// Take a screenshot in a second
+
+			if ( CVAR_GET_FLOAT( "cl_autorecord_stop") != 0 && CVAR_GET_FLOAT( "cl_autorecord") != 0)
```
YaLTeR: Not heavily related to `cl_autorecord`, no need to check if it's enabled.
```suggestion
			if ( CVAR_GET_FLOAT( "cl_autorecord_stop") != 0)
```

cl_dll/hud_countdown.cpp
```diff
@@ -19,6 +19,7 @@ int CHudCountdown::Init()
 	name2[ARRAYSIZE(name2) - 1] = '\0';
 
 	cl_autorecord = CVAR_CREATE("cl_autorecord", "0", FCVAR_ARCHIVE);
+	cl_autorecord_stop = CVAR_CREATE("cl_autorecord_stop", "1", FCVAR_ARCHIVE);
```
YaLTeR: Don't default this to enabled as it changes existing behavior.
```suggestion
	cl_autorecord_stop = CVAR_CREATE("cl_autorecord_stop", "0", FCVAR_ARCHIVE);
```

cl_dll/hud_countdown.h
```diff
@@ -10,6 +10,7 @@ class CHudCountdown : public CHudBase
 	char name2[32];
 
 	cvar_t* cl_autorecord;
+	cvar_t* cl_autorecord_stop;
```
YaLTeR: 1. Since it's not related to countdown, but rather `CHud` itself, it's better to put it into `CHud` (look for `m_pCvarColor` in `hud.h` and `hud.cpp` and put around there).

2. When you do this you'll be able to access the `cl_autorecord_stop` variable directly, so change `CVAR_GET_FLOAT()` in `CHud::Redraw()` to direct access as it's faster and less error-prone (`cl_autorecord_stop->value > 0.0f`).

3. `cl_autorecord_stop` doesn't really make sense as a name, it's a general demo auto-stop behavior which can be used apart from auto-record. Maybe `cl_autostop`?

cl_dll/hud.h
```diff
@@ -590,6 +590,7 @@ class CHud
 	int m_iFontHeight;
 
 	cvar_t* m_pCvarColor;
+	cvar_t* cl_autostop;
```
YaLTeR: Could you move this to right below `m_pCvarDrawDeathNoticesAlways` and name it `m_pCvarAutostop`? Right now it sits right in the middle of two `hud_color`-related things and its name is inconsistent with the others.

YaLTeR:
Thanks!

PR #63, opened 2019-12-07
[appveyor.yml] Use proper configuration matrix

Margen67:


YaLTeR:
Hm?

Margen67:
oops

YaLTeR:
Is this stated in the docs or something? Could you link?

Margen67:
https://www.appveyor.com/docs/appveyor-yml/
```
# to add several configurations to build matrix:
#configuration:
#  - Debug
#  - Release
```

YaLTeR:
Thing is, this isn't building a regular .sln, but rather passing the environment variable to CMake. I suspect your change won't work.

YaLTeR:
Even if this somehow works, I prefer the current approach because I'm explicitly using the `CONFIGURATION` environment variable to the `cmake` command, so I'd rather explicitly set it as an environment variable.

Margen67:
I'm pretty sure this method should still have the environment variable set.

PR #68, opened 2020-03-26
Add viewroll and viewmodel bob tilt

sabian-heatwave:
Viewroll and viewmodel bob tilt was loved by many players at Half-Life's release but was removed from the engine in version 1.1.1.0. I think that adding this would attract more players to AG as it brings a removed old-school feature back.

Video demonstration:
https://www.youtube.com/watch?v=Yf2YzRkdVOU

### DIFF
--- a/cl_dll/view.cpp
+++ b/cl_dll/view.cpp
@@ -411,7 +411,7 @@ void V_CalcViewRoll ( struct ref_params_s *pparams )
 	viewentity = gEngfuncs.GetEntityByIndex( pparams->viewentity );
 	if ( !viewentity )
 		return;
-	pparams->viewangles[ROLL] = V_CalcRoll (pparams->viewangles, pparams->simvel, cl_viewrollangle->value, cl_viewrollspeed->value ) * 4;
+	pparams->viewangles[ROLL] += V_CalcRoll (pparams->viewangles, pparams->simvel, cl_viewrollangle->value, cl_viewrollspeed->value ) * 4;
 
 	side = V_CalcRoll ( viewentity->angles, pparams->simvel, pparams->movevars->rollangle, pparams->movevars->rollspeed );
 
--- a/cl_dll/view.cpp
+++ b/cl_dll/view.cpp
@@ -401,6 +401,8 @@ V_CalcViewRoll
 Roll is induced by movement and damage
 ==============
 */
+extern cvar_t *cl_viewrollangle;
+extern cvar_t *cl_viewrollspeed;
 void V_CalcViewRoll ( struct ref_params_s *pparams )
 {
 	float		side;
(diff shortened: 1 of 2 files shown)


cl_dll/view.cpp
```diff
@@ -661,6 +664,7 @@ void V_CalcNormalRefdef ( struct ref_params_s *pparams )
 	view->angles[YAW]   -= bob * 0.5;
 	view->angles[ROLL]  -= bob * 1;
 	view->angles[PITCH] -= bob * 0.3;
+	VectorCopy(view->angles, view->curstate.angles);
```
YaLTeR: What is this for? Seems unrelated to the change.

sabian-heatwave: It makes the viewmodel bob go corner-to-corner as shown in the video instead of just back-and-forth, I think that's why it's commented "throw in a little tilt"

sabian-heatwave:
> Sure, but it should be disabled by default.

@YaLTeR I will make that change now.

cl_dll/view.cpp
YaLTeR: I'm asking specifically about your VectorCopy addition.

sabian-heatwave: The VectorCopy actually makes the bob go corner-to-corner (I think.)

YaLTeR: Could you remove it and check? I'd rather not introduce extra lines of code with little understanding of what they do.

cl_dll/view.cpp
```diff
@@ -409,6 +411,7 @@ void V_CalcViewRoll ( struct ref_params_s *pparams )
 	viewentity = gEngfuncs.GetEntityByIndex( pparams->viewentity );
 	if ( !viewentity )
 		return;
+	pparams->viewangles[ROLL] = V_CalcRoll (pparams->viewangles, pparams->simvel, cl_viewrollangle->value, cl_viewrollspeed->value ) * 4;
```
YaLTeR: Should this be a += so as to not override the roll angle?

cl_dll/view.cpp
sabian-heatwave: I already did that with [another SDK](https://github.com/thephoenixprojctsoftware/zamnhl). Removing it does not allow the bob to go corner-to-corner, the bob goes back-and-forth.

cl_dll/view.cpp
sabian-heatwave: I've actually never tried that before.. will do now.

sabian-heatwave: Adding `+=` made no in-game difference

YaLTeR: Probably because there was no viewangle roll. I suggest then changing it to += because the code right below also does += and not =.

cl_dll/view.cpp
YaLTeR: I checked myself and it does indeed make that difference. Alright.

YaLTeR:
Thanks!

PR #70, opened 2020-03-27
Change HUD colour to green to match AG branding

sabian-heatwave:
The AG logo is green (the modern one at least) and I have tried the HUD in-game and I think it goes very well, especially with the built-in maps such as `stalkx`. If you don't believe me, you should try it yourself, because I think it goes very well
-S


TheSmiley47:
If you want to change the color of hud for yourself, you can just write **hud_color "R G B"** in console, why is it necessary to change default color of hud?

sabian-heatwave:
> If you want to change the color of hud for yourself, you can just write **hud_color "R G B"** in console, why is it necessary to change default color of hud?

You can do that?!

TheSmiley47:
yes lol

mxpph:
Yes. It might be an idea to check what's already been implemented before implementing it yourself...

sabian-heatwave:
It doesn't work for me... I'll close this anyway for players who can.

mxpph:
Try hud_color "0 255 0"

fireblizzard:
I'm not sure what to think about the idea. I believe you should make a poll where most if not all of the AG community takes part.
About the change itself, I don't like these things:
- You're changing a value without changing its respective comment next to it, which should have been changed to //0,150,23
- You're changing a value that renders the define name useless, as you kept it as _RGB_YELLOWISH_ when it's really more greenish.
- Instead of changing that define, just make another one and change the usages of it to your new define.

That might serve you to learn some better practices. The third point makes the other 2 useless to point out, but I'm just trying to help you understand the thinking process.

sabian-heatwave:
I've made the changes you've suggested @fireblizzard, thank you :)

YaLTeR:
Sorry, but the default HUD color is not going to change. It's something everyone is used to and likes.

If you want to change your HUD color, please use the `hud_color` console variable as noted above.

PR #71, opened 2020-03-29
Add GitHub Actions CI for Windows

Margen67:
Windows only for now since macOS and Linux fail: https://github.com/Margen67/OpenAG/runs/461881133

Only has artifacts for the `v141_xp` toolset.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,44 @@
+name: CI
+
+on:
+  push:
+    paths-ignore:
+      - .travis.yml
+      - appveyor.yml
+      - filecopy.bat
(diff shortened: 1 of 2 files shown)


YaLTeR:
Awesome! For Linux you want to `sudo apt-get install rapidjson-dev mesa-common-dev`. Also could you add a VS 2019 build too (just its default toolset)?

YaLTeR:
Also pull master to try again on macOS, I removed that malloc.h line, seemed unused.

Margen67:
@YaLTeR Linux succeeds, but macOS still fails: https://github.com/Margen67/OpenAG/runs/542886541

PR #73, opened 2020-04-07
Add CVar to enable & disable old viewmodel bob tilt

barney-heatwave:


.gitignore
```diff
@@ -4,3 +4,4 @@ linux/release
 build
 compile_commands.json
 thirdparty
+.vs
```
barney-heatwave, line 7: The reason I've added this is that my Visual Studio always makes a `.vs` folder in the repository.

cl_dll/view.cpp
```diff
@@ -90,6 +90,7 @@ cvar_t	*cl_bob;
 cvar_t	*cl_bobup;
 cvar_t	*cl_waterdist;
 cvar_t	*cl_chasedist;
+cvar_t* cl_wonbob;
```
YaLTeR, line 93: I suggest calling it `bobwon` rather than `wonbob` so it shows up when you type `cl_bob` among the other variables. Even better, something like `bobtilt` so it's actually a descriptive name.

YaLTeR:
Also, why is this cvar even necessary?

cl_dll/view.cpp
barney-heatwave: Changes made in commit 296734c

barney-heatwave:
> Also, why is this cvar even necessary? Either leave this in without another cvar, or remove it altogether.

@YaLTeR I've put the default CVar value to 0 just in case that helps.
The reason why I've done this is that people might not want the tilted bob and some people like me might play with bob still on, and sometimes I like to have the option to turn it on and off. This CVar is just a suggestion to add to OpenAG. Most players may not notice it after the most recent build since the default is 0 but it is nice to have the option in my opinion. 

If you don't want the CVar I can close this pull request or I can remove bobtilt.

YaLTeR:
> people might not want the tilted bob

Let's wait till someone actually complains. Until then, closing.

PR #74, opened 2020-05-01
CI improvements

Margen67:


YaLTeR:
It seems to use clang 9 on all four builds:
```
-- The C compiler identification is Clang 9.0.0
-- The CXX compiler identification is Clang 9.0.0
```
You don't seem to be installing different clang versions? Also I don't think installing the same gcc version as clang is required.

Margen67:
> 
> 
> It seems to use clang 9 on all four builds:
> 
> ```
> -- The C compiler identification is Clang 9.0.0
> -- The CXX compiler identification is Clang 9.0.0
> ```
> 
> You don't seem to be installing different clang versions? Also I don't think installing the same gcc version as clang is required.

I forgot to put the clang version in the compiler environment variable.

.github/workflows/CI.yml
```diff
@@ -27,7 +33,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        submodules: recursive
+        submodules: true
```
YaLTeR, line 36: I think it's best to always do `recursive` unless there's a specific reason not to. It's almost always what you want. Even if there are no recursive submodules now, it better just work if any appear later on.

```suggestion
        submodules: recursive
```

.github/workflows/CI.yml
```diff
@@ -50,15 +56,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gcc-ver: [6, 7, 8]
+        gcc-ver: [7, 8, 9, 10]
```
YaLTeR, line 59: I'd rather keep testing on the oldest supported compiler until I explicitly decide to break backwards compatibility.

```suggestion
        gcc-ver: [6, 7, 8, 9, 10]
```

.github/workflows/CI.yml
```diff
@@ -50,15 +56,14 @@ jobs:
...
         configuration: [Release, Debug]
     steps:
     - uses: actions/checkout@v2
       with:
-        submodules: recursive
+        submodules: true
```
YaLTeR, line 64:
```suggestion
        submodules: recursive
```

.github/workflows/CI.yml
```diff
@@ -67,39 +72,38 @@ jobs:
...
     - uses: actions/upload-artifact@v1
-      if: matrix.gcc-ver == '8'
+      if: matrix.gcc-ver == '10'
       with:
         name: openag-${{ runner.os }}-${{ matrix.configuration }}
         path: build/client.so
 
   build-linux-clang:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04 # TODO: Change this to ubuntu-latest once it becomes 20.04.
```
YaLTeR, line 83: I think it's fine to just leave it as `ubuntu-20.04`. We're testing on a wide set of compilers rather than for "compiles on the latest Ubuntu".

```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/CI.yml
```diff
@@ -67,39 +72,38 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
+        clang-ver: [9, 10]
         configuration: [Release, Debug]
     steps:
     - uses: actions/checkout@v2
       with:
-        submodules: recursive
+        submodules: true
```
YaLTeR, line 95:
```suggestion
        submodules: recursive
```

.github/workflows/CI.yml
```diff
@@ -111,17 +115,16 @@ jobs:
   #  steps:
   #  - uses: actions/checkout@v2
   #    with:
-  #      submodules: recursive
+  #      submodules: true
```
YaLTeR, line 118:
```suggestion
  #      submodules: recursive
```

.github/workflows/CI.yml
```diff
@@ -3,22 +3,28 @@ name: CI
 on:
   push:
     paths-ignore:
-      - '.travis.yml'
-      - 'filecopy.bat'
+      - '.github/*'
+      - '.github/*_TEMPLATE/**'
```
YaLTeR, line 7: What's this `_TEMPLATE` folder?

YaLTeR (changes requested):

.github/workflows/CI.yml
Margen67: It's for issue/pull request templates.

.github/workflows/CI.yml
```diff
@@ -3,22 +3,29 @@ name: CI
...
+      - '*.bat'
+      - '*.yml'
```
YaLTeR, line 9: Doesn't that block github's own CI .yml changes?

.github/workflows/CI.yml
```diff
@@ -3,22 +3,29 @@ name: CI
...
+      - '*.md'
       - 'LICENSE*'
-      - 'README.md'
+  workflow_dispatch:
```
YaLTeR, line 20: What's this for?

.github/workflows/CI.yml
```diff
@@ -36,63 +43,75 @@ jobs:
...
         path: build\${{ matrix.configuration }}\client.dll
+        if-no-files-found: error
 
   build-linux-gcc:
-    runs-on: ubuntu-18.04
-    env:
-      CC: gcc-${{ matrix.gcc-ver }}
-      CXX: g++-${{ matrix.gcc-ver }}
+    name: build-linux (gcc-${{ matrix.gcc-ver }}, ${{ matrix.configuration }})
+    runs-on: ubuntu-latest
```
YaLTeR, line 55: This needs to be `ubuntu-18.04` for two reasons. First, `-latest` automatically breaks CI when it changes, I'm not a big fan of that. Second, the built `client.so` links to symbols from the distribution glibc, `ubuntu-20.04` has a newer glibc, so compiling on `ubuntu-20.04` makes `cilent.so` incompatible with Ubuntu 18.04 and other slightly outdated distributions.

.github/workflows/CI.yml
```diff
@@ -36,63 +43,75 @@ jobs:
...
+        if-no-files-found: error
 
   build-linux-clang:
-    runs-on: ubuntu-18.04
-    env:
-      compiler: clang
-      CC: clang
-      CXX: clang++
+    name: build-linux (clang-${{ matrix.clang-ver }}, ${{ matrix.configuration }})
+    runs-on: ubuntu-latest
```
YaLTeR, line 94: Also needs to be `ubuntu-18.04` or `20.04` since artifacts aren't uploaded from this.

YaLTeR (changes requested):

.github/workflows/CI.yml
Margen67: It only does it for the root directory. It would need to be `'**/*.yml'` to be recursive.

.github/workflows/CI.yml
Margen67: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

.github/workflows/CI.yml
```diff
@@ -3,100 +3,111 @@ name: CI
...
+        with:
+          submodules: recursive
+      - name: CMake generate
+        run: cmake -T ${{ matrix.toolset }} -A Win32 -B build
+      - name: Build
+        working-directory: build
+        run: cmake --build . -j $env:NUMBER_OF_PROCESSORS --config ${{ matrix.configuration }}
+      - uses: actions/upload-artifact@v2
+        with:
+          name: openag-${{ runner.os }}-${{ matrix.toolset }}-${{ matrix.configuration }}
```
YaLTeR, line 45: We don't really need other binaries, only the v141_xp one.

YaLTeR:
Cool, thanks a lot!

PR #76, opened 2020-05-25
Add a cvar to fix viewheight on newer HLSDK

fireblizzard:
Fixes the height of the first person cam when spectating in games that use the latest HLSDK, like HL.
When playing HL with OpenAG you would have the cam 28 units higher than it should be in First Person spectator mode, which is a bit annoying.

`cl_viewheight_mode 0` - normal AG viewheight
`cl_viewheight_mode 1` - fixed viewheight for HL

### DIFF
--- a/cl_dll/hud.cpp
+++ b/cl_dll/hud.cpp
@@ -482,6 +482,7 @@ void CHud :: Init( void )
 	m_pCvarDraw = CVAR_CREATE( "hud_draw", "1", FCVAR_ARCHIVE );
 	m_pCvarDrawDeathNoticesAlways = CVAR_CREATE( "cl_draw_deathnotices_always", "0", FCVAR_ARCHIVE );
 	m_pCvarAutostop = CVAR_CREATE("cl_autostop", "0", FCVAR_ARCHIVE);
+	m_pCvarViewheightMode = CVAR_CREATE("cl_viewheight_mode", "0", FCVAR_ARCHIVE);
 	m_pCvarColor = CVAR_CREATE( "hud_color", "", FCVAR_ARCHIVE );
 	cl_lw = gEngfuncs.pfnGetCvarPointer( "cl_lw" );
 
--- a/cl_dll/hud.h
+++ b/cl_dll/hud.h
@@ -587,6 +587,7 @@ public:
 	cvar_t	*m_pCvarDraw;
 	cvar_t	*m_pCvarDrawDeathNoticesAlways;
 	cvar_t	*m_pCvarAutostop;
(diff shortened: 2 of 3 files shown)


fireblizzard:
I tried to the imitate surrounding code style. The condition line may be too long.

YaLTeR:
Thanks!

rtxa:
@YaLTeR @fireblizzard I would like to suggest to add some check to do this automatically, because most players will not know about this or how to configure it. If the server has sv_ag_version 6.3 or 6.6, keep the viewheight in 0, if the cvar isn't present, then leaved in 1. Maybe I can do the pullrequest, but I would like to know if it's a good idea. Thanks.

fireblizzard:
> @YaLTeR @fireblizzard I would like to suggest to add some check to do this automatically, because most players will not know about this or how to configure it. If the server has sv_ag_version 6.3 or 6.6, keep the viewheight in 0, if the cvar isn't present, then leaved in 1. Maybe I can do the pullrequest, but I would like to know if it's a good idea. Thanks.

It would be better to have it automated.

But I don't know what would be the criteria for deciding whether to fix it. You know that you would NOT have to fix the viewheight if there's `sv_ag_version`, just leave it as is. But how do you know if you have to actually fix it? Is there anything that tells you if the HLSDK version that the game is using is updated enough to have the Observer code serverside, messing with viewheight? Can we assume that any multiplayer mod (as I don't think there's spectator mode for singleplayer) should be using such an updated HLSDK, so always fix the viewheight if `sv_ag_version` is missing?

rtxa:
I think the probability to use this client in another mod besides HL or AG it's almost none. In any case, a cvar `cl_viewheight_auto` should be added in case the auto-detection fails.

YaLTeR:
> a cvar cl_viewheight_auto should be added in case the auto-detection fails

I think that, considering where OpenAG is used, an automatic check is only worth it if it cannot really fail, otherwise it's not worth the effort.

rtxa:
> 
> 
> > a cvar cl_viewheight_auto should be added in case the auto-detection fails
> 
> I think that, considering where OpenAG is used, an automatic check is only worth it if it cannot really fail, otherwise it's not worth the effort.

mm, it's not a big deal to check ag version, and the cases where it can fails can be very strange, what matters is that need to works fine in ag

YaLTeR:
It works fine in AG as it is because the default value is to not do anything.

rtxa:
> 
> 
> It works fine in AG as it is because the default value is to not do anything.

aha, what's your point?

PR #77, opened 2020-05-26
Multilayer overviews

fireblizzard:
Valve added some support for multiple overview layers, but it seems that at the end they decided not to support them. So this is mostly refactoring of overview code and completing the support.

Some background on overviews:
- An overview is the top-down view of the map that you can see from spectator mode when pressing `+use`. You can rotate it by moving the mouse, and height (Z) is only taken into account for positioning player indicators or yourself, with a line going from the indicator to the overview image where you can see at which height players are depending on how long is the line. [Example overview image](https://imgur.com/a/3he7Ice).
- An overview is defined in a txt file located in the _overviews_ directory, with the same name as the corresponding map. [Example overview file (multilayered)](https://pastebin.com/LmQ33xnf).
- It has some properties like `zoom`, `origin`, `rotated`, and then you have the layers, that have a couple properties like `image` and `height`. They're more or less self-explanatory if you have seen overviews ingame. `height` is the minimum Z for a layer, and originally it was only used for positioning yourself higher or lower in the overview, but now it's also used to change the displayed overview image. So, for example, right when your origin (your cam's) goes below the current layer's `height`, it will switch to the next (lower) layer, if any.

This PR doesn't change the way these properties are used, so backwards compatibility is kept, ie. existing overviews will continue working the same way.

Here's a demonstration video of multilayered overviews. You can see how it switches the overview image based on your position (Z):
https://www.youtube.com/watch?v=3VJv8u2QOjo

### DIFF
--- a/cl_dll/hud_spectator.cpp
+++ b/cl_dll/hud_spectator.cpp
@@ -594,7 +594,7 @@ int CHudSpectator::VidInit()
 	m_hsprPlayerBlue	= SPR_Load("sprites/iplayerblue.spr");
 	m_hsprPlayerRed		= SPR_Load("sprites/iplayerred.spr");
 	m_hsprPlayerDead	= SPR_Load("sprites/iplayerdead.spr");
-	m_hsprUnkownMap		= SPR_Load("sprites/tile.spr");
+	m_hsprUnknownMap	= SPR_Load("sprites/tile.spr");
 	m_hsprBeam			= SPR_Load("sprites/laserbeam.spr");
 	m_hsprCamera		= SPR_Load("sprites/camera.spr");
 	m_hCrosshair		= SPR_Load("sprites/crosshairs.spr");
@@ -1279,18 +1279,8 @@ bool CHudSpectator::ParseOverviewFile( )
 	float height;
 	
 	char *pfile  = NULL;
-
-	memset( &m_OverviewData, 0, sizeof(m_OverviewData));
-
-	// fill in standrd values
-	m_OverviewData.insetWindowX = 4;	// upper left corner
-	m_OverviewData.insetWindowY = 4;
-	m_OverviewData.insetWindowHeight = 180;
-	m_OverviewData.insetWindowWidth = 240;
-	m_OverviewData.origin[0] = 0.0f;
-	m_OverviewData.origin[1] = 0.0f;
-	m_OverviewData.origin[2] = 0.0f;
-	m_OverviewData.zoom	= 1.0f;
+	
+	m_OverviewData = Overview();
 	m_OverviewData.map = gEngfuncs.pfnGetLevelName();
 
 	const auto mapLen = m_OverviewData.map.size();
@@ -1305,7 +1295,7 @@ bool CHudSpectator::ParseOverviewFile( )
 
 	if (!pfile)
 	{
-		gEngfuncs.Con_DPrintf("Couldn't open file %s. Using default values for overiew mode.\n", filename.c_str() );
+		gEngfuncs.Con_DPrintf("Couldn't open file %s. Using default values for overview mode.\n", filename.c_str() );
 		return false;
 	}
 	
@@ -1364,7 +1354,7 @@ bool CHudSpectator::ParseOverviewFile( )
 				}
 				else
 				{
-					gEngfuncs.Con_Printf("Error parsing overview file %s. (%s unkown)\n", filename.c_str(), token );
+					gEngfuncs.Con_Printf("Error parsing overview file %s. (%s unknown)\n", filename.c_str(), token );
 					return false;
 				}
 
@@ -1406,7 +1396,7 @@ bool CHudSpectator::ParseOverviewFile( )
 				}
 				else
 				{
-					gEngfuncs.Con_Printf("Error parsing overview file %s. (%s unkown)\n", filename.c_str(), token );
+					gEngfuncs.Con_Printf("Error parsing overview file %s. (%s unknown)\n", filename.c_str(), token );
 					return false;
 				}
 
@@ -1432,7 +1422,7 @@ void CHudSpectator::LoadMapSprites()
 	for (auto& layer : m_OverviewData.layers)
 	{
 		if (layer.imagePath.empty())
-			layer.mapSprite = (struct model_s*) gEngfuncs.GetSpritePointer(m_hsprUnkownMap);
+			layer.mapSprite = (struct model_s*) gEngfuncs.GetSpritePointer(m_hsprUnknownMap);
 		else
 			layer.mapSprite = gEngfuncs.LoadMapSprite(layer.imagePath.c_str());
 	}
@@ -1447,12 +1437,13 @@ void CHudSpectator::DrawOverviewLayer()
 	GetCameraView(camOrigin, camAngles);
 
 	const auto currLayer = GetCurrentLayer(camOrigin);
-	const auto hasMapImage = currLayer.mapSprite == (struct model_s*)gEngfuncs.GetSpritePointer(m_hsprUnkownMap) ? false : true;
+	const auto unknown_map = (struct model_s*)gEngfuncs.GetSpritePointer(m_hsprUnknownMap);
+	const auto hasMapImage = (currLayer.mapSprite != unknown_map);
 
 	if ( hasMapImage)
 	{
 		i = currLayer.mapSprite->numframes / (4*3);
-		i = (int) sqrt((float)i);
+		i = sqrt((float)i);
 		xTiles = i*4;
 		yTiles = i*3;
 	}
@@ -1945,7 +1936,7 @@ void CHudSpectator::Reset()
 		ParseOverviewFile();
 
 		// A default layer for dummy overview when there's no overview file
-		if (m_OverviewData.layers.size() == 0)
+		if (m_OverviewData.layers.empty())
 			m_OverviewData.layers.push_back(OverviewLayer());
 
 		LoadMapSprites();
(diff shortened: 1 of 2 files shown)


cl_dll/hud_spectator.cpp
```diff
@@ -1255,8 +1273,8 @@ bool CHudSpectator::IsActivePlayer(cl_entity_t * ent)
 
 bool CHudSpectator::ParseOverviewFile( )
 {
-	char filename[255];
-	char levelname[255];
+	std::string filename;
+	std::string levelname;
 	char token[1024];
 	float height;
```
YaLTeR, line 1280: `memset( &m_OverviewData, 0, sizeof(m_OverviewData));` just below this line is UB, especially so since `Overview` is non-POD. Remove it and move the default value setting from here to the default constructor, and just call the default constructor (`m_OverviewData = Overview();`).

cl_dll/hud_spectator.cpp
```diff
@@ -1417,27 +1429,30 @@ bool CHudSpectator::ParseOverviewFile( )
...
 	float screenaspect, xs, ys, xStep, yStep, x,y,z;
 	int ix,iy,i,xTiles,yTiles,frame;
 
-	qboolean	hasMapImage = m_MapSprite?TRUE:FALSE;
-	model_t *   dummySprite = (struct model_s *)gEngfuncs.GetSpritePointer( m_hsprUnkownMap);
+	Vector camOrigin, camAngles;
+	GetCameraView(camOrigin, camAngles);
+
+	const auto currLayer = GetCurrentLayer(camOrigin);
+	const auto hasMapImage = currLayer.mapSprite == (struct model_s*)gEngfuncs.GetSpritePointer(m_hsprUnkownMap) ? false : true;
```
YaLTeR, line 1450:
```suggestion
	const auto unknown_map = (struct model_s*)gEngfuncs.GetSpritePointer(m_hsprUnkownMap);
	const auto hasMapImage = (currLayer.mapSprite != unknown_map);
```

cl_dll/hud_spectator.cpp
```diff
@@ -1417,27 +1429,30 @@ bool CHudSpectator::ParseOverviewFile( )
...
 
 	if ( hasMapImage)
 	{
-		i = m_MapSprite->numframes / (4*3);
-		i = sqrt((float)i);
+		i = currLayer.mapSprite->numframes / (4*3);
+		i = (int) sqrt((float)i);
```
YaLTeR, line 1455: C++ implicitly converts float to int.
```suggestion
		i = sqrt((float)i);
```

cl_dll/hud_spectator.cpp
```diff
@@ -1943,10 +1939,15 @@ int CHudSpectator::ToggleInset(bool allowOff)
...
 {
 	// Reset HUD
-	if ( strcmp( m_OverviewData.map, gEngfuncs.pfnGetLevelName() ) )
+	if ( strcmp( m_OverviewData.map.c_str(), gEngfuncs.pfnGetLevelName() ) )
 	{
 		// update level overview if level changed
 		ParseOverviewFile();
+
+		// A default layer for dummy overview when there's no overview file
+		if (m_OverviewData.layers.size() == 0)
```
YaLTeR, line 1948:
```suggestion
		if (m_OverviewData.layers.empty())
```

cl_dll/hud_spectator.cpp
```diff
@@ -1988,3 +1989,33 @@ void CHudSpectator::InitHUDData()
 	gHUD.m_iFOV =  CVAR_GET_FLOAT("default_fov");
 }
 
+CHudSpectator::OverviewLayer CHudSpectator::GetCurrentLayer(Vector playerPos)
+{
+	// Get the layer with highest Z, that will be the default layer,
+	// or it will be the dummy oneif an overview file doesn't exist for this map
```
YaLTeR, line 1995:
```suggestion
	// or it will be the dummy one if an overview file doesn't exist for this map
```

cl_dll/hud_spectator.cpp
```diff
@@ -1988,3 +1989,33 @@ void CHudSpectator::InitHUDData()
...
+	OverviewLayer result = GetHighestLayer();
+
+	// Get the overview with highest Z that is below player Z position too
+	for (const auto& layer : m_OverviewData.layers)
+	{
+		if (playerPos.z < layer.z && layer.z < result.z)
```
YaLTeR, line 2001: This gets the overview with the lowest Z that is above player Z position, directly contrary to the comment

cl_dll/hud_spectator.cpp
```diff
@@ -1988,3 +1989,33 @@ void CHudSpectator::InitHUDData()
...
+		{
+			result = layer;
+		}
+	}
+	return result;
+}
```
YaLTeR, line 2007:
```suggestion
}

```

cl_dll/hud_spectator.h
```diff
@@ -67,6 +52,31 @@ typedef struct cameraWayPoint_s
...
+		std::string		imagePath;
+		float			z;
+		model_s*		mapSprite;
+		OverviewLayer()
+			: imagePath("")
+			, z(0.0f)
+			, mapSprite(nullptr)
+		{
+		}
+	};
```
YaLTeR, line 66:
```suggestion
	};
	
```

YaLTeR:
Also change unkown to unknown pls

cl_dll/hud_spectator.cpp
fireblizzard: Thanks, I'll update the comment. It was for a different approach that I had and forgot to change the comment accordingly.

fireblizzard:
I didn't initialize `map` and `layers` in the constructor, because I've [read](https://stackoverflow.com/a/11725466) that it's not necessary, but I'm not sure if not making them explicit goes against this project's code style or yours. Then, I wanted to pass the `gEngfuncs.pfnGetLevelName()` as parameter to fill the `map` string but it would say this:

> 'CHudSpectator::m_OverviewData' has either no appropriate default constructor or overload resolution was ambiguous

And initializing `map` with that directly, without being passed as parameter, would crash the game on startup. So I just set the `map` afterwards.

The check corresponding to `CI / build-linux-gcc (8, Release)` has failed in the Setup step, I'm not sure if it's latest commit's fault, looks like it failed to retrieve some package (_jdk1.8_).

YaLTeR:
> I didn't initialize `map` and `layers` in the constructor, because I've [read](https://stackoverflow.com/a/11725466) that it's not necessary

Yeah, for types with default constructors it's not necessary.

> but I'm not sure if not making them explicit goes against this project's code style or yours.

Doesn't go against any project style.

> Then, I wanted to pass the `gEngfuncs.pfnGetLevelName()` as parameter to fill the `map` string but it would say this:
> 
> > 'CHudSpectator::m_OverviewData' has either no appropriate default constructor or overload resolution was ambiguous
> 
> And initializing `map` with that directly, without being passed as parameter, would crash the game on startup. So I just set the `map` afterwards.

Could you show code that you tried for both of these? I'll tell you what was wrong.

> The check corresponding to `CI / build-linux-gcc (8, Release)` has failed in the Setup step, I'm not sure if it's latest commit's fault, looks like it failed to retrieve some package (_jdk1.8_).

Yeah, CI being CI, happens.

fireblizzard:
Sorry for the delay.

> Could you show code that you tried for both of these? I'll tell you what was wrong.

I tried this:
```
// hud_spectator.h # CHudSpectator

Overview(const std::string& map)
	: map(map)
	, origin(Vector(0, 0, 0))
	, zoom(1.0f)
	, rotated(false)
	, insetWindowX(4)
	, insetWindowY(4)
	, insetWindowHeight(180)
	, insetWindowWidth(240)
{
}
```
```
// hud_spectator.cpp # CHudSpectator::ParseOverviewFile

m_OverviewData = Overview(gEngfuncs.pfnGetLevelName());
```
Which fails to compile, with messages like this:
> note: 'CHudSpectator::CHudSpectator(void)': function was implicitly deleted because a data member 'CHudSpectator::m_OverviewData' has either no appropriate default constructor or overload resolution was ambiguous

Which I believe it's because `Overview  m_OverviewData` in `CHudSpectator` is trying to use a default constructor without parameters, so I have to either make a new constructor without parameters, or do `Overview  m_OverviewData = Overview("");`, or leave it as I commited.


The other thing I tried was initializing map with the call to gEngfuncs to get the map name, which I believe it fails because at that time `gEngfuncs` is not a thing yet, or some other thing that I can't imagine:
```
// hud_spectator.h # CHudSpectator

Overview()
	: map(gEngfuncs.pfnGetLevelName())
	, origin(Vector(0, 0, 0))
	, zoom(1.0f)
	, rotated(false)
	, insetWindowX(4)
	, insetWindowY(4)
	, insetWindowHeight(180)
	, insetWindowWidth(240)
{
}
```
Compiles, but fails right after starting the game. I see the main menu for an instant and then throws this error:
> could not load library <some_path>\steamapps\common\half-life\ag\cl_dlls\client.dll

YaLTeR:
Ah, that's right. What you could do is have both the no-argument constructor and the map name constructor. Or just leave it as is I guess.

fireblizzard:
I've done it like they say [here](https://stackoverflow.com/a/308318), but I'm not sure about what's the proper code style for the default constructor. If you have any suggestion I can continue doing any necessary change.

fireblizzard:
I think it's good to go. The pending requested changes are already done, the ones with a newline after braces. It's just not showing them as completed.

YaLTeR:
Thanks!

PR #78, opened 2020-06-04
Client: Add cvar to hide corpses and fix standing corpses

rtxa:
Cvar `cl_fix_standing_corpses ` fixes the issue with standing corpses leaved by players playing with high FPS.

Cvar `cl_hidecorpses `gives you the ability to hide the corpses in case they bother you. If there wasn't a fix for the standing corpses, it becomes useful to counter that issue.

### DIFF
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -49,12 +49,14 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 		break;
 	}
 
-	if (gHUD.m_pCvarFixStandingCorpses->value > 0 
-		&& (ent->player || ent->curstate.renderfx == kRenderFxDeadPlayer) && ent->curstate.framerate == 0)
-		ent->curstate.frame = 255;
-
+	// hide corpses option
 	if (gHUD.m_pCvarHideCorpses->value > 0 && ent->curstate.renderfx == kRenderFxDeadPlayer)
 		return 0;
+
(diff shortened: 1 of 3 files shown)


rtxa:
1. There's a case where the player is dead but he hasn't respawn yet (CopyBodyToQue() hasn't been called), so while in the time the player hasn't respawn, the death animation will still be unfinished. I'm not sure if this `ent->player` it's safe to use, I mean, that it will not interfere with anything that's not related to death animation, but in my testings, I didn't notice anything wrong. To make this more reliable, I need to check the pev->deadflag of that player, but I'm not sure if this is sended to the client.
2. Well, if the corpse is gonna be hidden, there's no need to set the death animation to the last frame, but anyway I'm gonna put the code for hide corpses before the fix for standing corpses.
3. I'm not sure, it can be 255.0 or 256.0, maybe you can help me to guess this from [StudioFrameAdvance()](https://github.com/ValveSoftware/halflife/blob/5d761709a31ce1e71488f2668321de05f791b405/dlls/animating.cpp#L62) code.
4. Ok.
5. Yeah.

rtxa:
I added the requested changes. I'll explain a little better the fix for standing corpses.
1. I check if the animation is done by checking if the framerate is equal to 0.
2. I check if the entity is a player or a corpse (bodyque classname) to set animation to his last frame.

I didn't have any issues in my testings and this works because I don't mess with animations that are already playing (An animation is only playing if his framerate is greater than zero).

YaLTeR:
Thanks!

PR #84, opened 2020-10-03
Add Unicode support to the scoreboard

tmp64:
This pull request adds support of Unicode to the scoreboard.

`UnicodeTextImage` class was added that replaces `vgui::TextImage` in the scoreboard. It uses VGUI2 ISurface interface (exposed by the engine) to draw Unicode chars and load fonts. If VGUI2 is not available (e.g. an engine update broke interface compatibility or game is running on Xash3D), it will fall back to `vgui::TextImage`. VGUI2 fonts also support anti-aliasing and it's enabled for large fonts (small fonts look blurry with AA).

To draw Unicode chars, UTF-8 text needs to be converted to a wide string (UTF-16 on Windows, UTF-32 on Linux and macOS). It is done using `ILocalize::ConvertANSIToUnicode` (exported by vgui2.dll). I couldn't use `std::mbstowcs` because it doesn't work on Windows.

A few cvars were added that control behavior of UnicodeTextImage.
- `vgui_unicode` - enables/disables unicode support and font rendering via VGUI2. If changed after the map was loaded, requires game restart.
- `vgui_unicode_aa` - enables/disables anti-aliasing of fonts. If changed after the map was loaded, requires game restart. Value of 2 enables AA for all fonts (even small ones).

Comparison of behavior (best viewed in full resolution)
openag_unicode

A few screenshots (best viewed in full resolution):

ncp_rockbhop_ez0000

crossfire0000

I tested font rendering on resolutions 640x480, 800x600, 1024x768, 1600x900 and 1920x1080 and didn't find any problems.

The pull request was tested on Linux and I don't see any reason it wouldn't work on macOS (if it's supported).

### DIFF
--- a/cl_dll/vgui_ScorePanel.cpp
+++ b/cl_dll/vgui_ScorePanel.cpp
@@ -102,6 +102,10 @@ ScorePanel::ScorePanel(int x,int y,int wide,int tall) : Panel(x,y,wide,tall)
 	Font *tfont = pSchemes->getFont(hTitleScheme);
 	Font *smallfont = pSchemes->getFont(hSmallScheme);
 
+	m_UFont = UnicodeTextImage::createFont("Arial", YRES(12), 300, gHUD.m_pCvarVGuiUnicodeAA->value);
+	m_UTitleFont = UnicodeTextImage::createFont("Arial", tfont->getTall(), 700, gHUD.m_pCvarVGuiUnicodeAA->value);
+	m_USmallFont = UnicodeTextImage::createFont("Arial", smallfont->getTall(), 400, gHUD.m_pCvarVGuiUnicodeAA->value);
+
 	setBgColor(0, 0, 0, 96);
 	m_pCurrentHighlightLabel = NULL;
 	m_iHighlightRow = -1;
@@ -168,7 +172,7 @@ ScorePanel::ScorePanel(int x,int y,int wide,int tall) : Panel(x,y,wide,tall)
 
 		m_HeaderLabels[i].setBgColor(0,0,0,255);
 		m_HeaderLabels[i].setFgColor(Scheme::sc_primary1);
-		m_HeaderLabels[i].setFont(smallfont);
+		m_HeaderLabels[i].setFont(m_USmallFont, smallfont);
 		m_HeaderLabels[i].setContentAlignment(g_ColumnInfo[i].m_Alignment);
 
 		int yres = 12;
@@ -612,7 +616,7 @@ void ScorePanel::FillGrid()
 			pLabel->setVisible(true);
 			pLabel->setText2("");
 			pLabel->setImage(NULL);
-			pLabel->setFont(sfont);
+			pLabel->setFont(m_UFont, sfont);
 			pLabel->setTextOffset(0, 0);
 			
 			int rowheight = 13;
@@ -655,7 +659,7 @@ void ScorePanel::FillGrid()
 					rowheight = YRES(rowheight);
 				}
 				pLabel->setSize(pLabel->getWide(), rowheight);
-				pLabel->setFont(tfont);
+				pLabel->setFont(m_UTitleFont, tfont);
 
 				pGridRow->SetRowUnderline(	0,
 											true,
@@ -677,7 +681,7 @@ void ScorePanel::FillGrid()
 					rowheight = YRES(rowheight);
 				}
 				pLabel->setSize(pLabel->getWide(), rowheight);
-				pLabel->setFont(tfont);
+				pLabel->setFont(m_UTitleFont, tfont);
 
 				pGridRow->SetRowUnderline(0, true, YRES(3), 100, 100, 100, 0);
 
@@ -768,7 +772,7 @@ void ScorePanel::FillGrid()
 						}
 
 						pLabel->setText2(sz2);
-						pLabel->setFont2(smallfont);
+						pLabel->setFont2(m_USmallFont, smallfont);
 					}
 					break;
 				case COLUMN_VOICE:
--- a/cl_dll/vgui_ScorePanel.h
+++ b/cl_dll/vgui_ScorePanel.h
@@ -15,6 +15,7 @@
 #include<VGUI_Label.h>
 #include<VGUI_TextImage.h>
 #include "../game_shared/vgui_listbox.h"
+#include "vgui_UnicodeTextImage.h"
 
 #include <ctype.h>
 
@@ -40,8 +41,8 @@ class CTextImage2 : public Image
 public:
 	CTextImage2()
 	{
-		_image[0] = new TextImage("");
-		_image[1] = new TextImage("");
+		_image[0] = new UnicodeTextImage();
+		_image[1] = new UnicodeTextImage();
 	}
 
 	~CTextImage2()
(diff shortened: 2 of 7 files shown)


YaLTeR:
That's cool. Before I do an in-depth review, two things:
1. > A few cvars were added that control behavior of UnicodeTextImage.
    >
    > - `vgui_unicode` - enables/disables unicode support and font rendering via VGUI2. If changed after the map was loaded, requires game restart.
    > - `vgui_unicode_aa` - enables/disables anti-aliasing of fonts. If changed after the map was loaded, requires game restart. Value of 2 enables AA for all fonts (even small ones).

    Why is this needed? Are there any drawbacks?
1. The second commit does not need to be separate from the first one.

tmp64:
1. I added them just in case something goes wrong or someone doesn't like that font looks differently (VGUI1 uses bitmap fonts). 
2. The development was done on Windows and then I tested it on Linux. You can squash-merge the pull request into one commit.

YaLTeR:
> I added them just in case something goes wrong or someone doesn't like that font looks differently (VGUI1 uses bitmap fonts).

Eh, I'd rather make it unconditional until somebody complains with a sufficient reason. Could you remove the cvars and default to enabled with AA?

tmp64:
Done. I also found a bug with interface loading error handling that I fixed.

cl_dll/vgui_UnicodeTextImage.cpp
```diff
@@ -0,0 +1,558 @@
...
+	// Try hardware engine (hw.dll)
+	GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, "hw.dll", &hw);
+	if (hw)
+	{
+		g_hEngineModule = reinterpret_cast<CSysModule *>(hw);
+	}
+	else
+	{
+		// Try software engine (sw.dll)
+		GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, "hw.dll", &sw);
```
YaLTeR, line 282: the classic copy-paste
```suggestion
		GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, "sw.dll", &sw);
```

cl_dll/vgui_UnicodeTextImage.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+	g_bSurfaceLoaded = true;
+}
+
+bool UnicodeTextImage::shouldFallback()
+{
+	return !g_bSurfaceLoaded;
+}
+
+UnicodeTextImage::HFont UnicodeTextImage::createFont(const char *fontName, int tall, int weight, bool antialias)
```
YaLTeR, line 354: The `antialias` argument can be removed now that the cvars are removed.

cl_dll/vgui_UnicodeTextImage.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+class ILocalize : public IBaseInterface
+{
+public:
+	virtual void Something1() = 0;
+	virtual void Something2() = 0;
+	virtual void Something3() = 0;
+
+	// Converts UTF-8 to WString
+	virtual int ConvertANSIToUnicode(const char *ansi, wchar_t *unicode, int unicodeBufferSizeInBytes) = 0;
```
YaLTeR, line 244: What happens if there's a 3+-byte character like an emoji and `wchar_t` is 16 bytes?

cl_dll/vgui_ScorePanel.cpp
```diff
@@ -102,6 +102,10 @@ ScorePanel::ScorePanel(int x,int y,int wide,int tall) : Panel(x,y,wide,tall)
 	Font *tfont = pSchemes->getFont(hTitleScheme);
 	Font *smallfont = pSchemes->getFont(hSmallScheme);
 
+	m_UFont = UnicodeTextImage::createFont("Arial", YRES(12), 300, true);
+	m_UTitleFont = UnicodeTextImage::createFont("Arial", tfont->getTall(), 700, true);
+	m_USmallFont = UnicodeTextImage::createFont("Arial", smallfont->getTall(), 400, true);
```
YaLTeR, line 107: Do these need to be freed in the destructor?

cl_dll/vgui_ScorePanel.h
```diff
@@ -149,26 +150,25 @@ class CLabelHeader : public Label
...
-
-		_gap += XRES(5);
+		int wide, tall;
+		_dualImage->GetImage(0)->getTextSize(wide, tall);
+		_gap = wide + XRES(5);
 	}
 
 	virtual void setText(const char* text)
 	{
+		// std::isspace assert-fails on UTF-8 chars
```
YaLTeR, line 160: How was it not an issue before?

YaLTeR (changes requested):

cl_dll/vgui_UnicodeTextImage.cpp
tmp64: VGUI2 only supports Basic Multilingual Plane (0000-FFFF). Such characters cannot be typed in the console or settings dialog.

cl_dll/vgui_ScorePanel.cpp
tmp64: No, all loaded fonts are freed by the engine's font manager on shutdown.

cl_dll/vgui_ScorePanel.h
tmp64: Assertions are removed from release builds and I guess no dev encountered a player with non-english chars in the name.

cl_dll/vgui_UnicodeTextImage.cpp
tmp64: I removed it.

cl_dll/vgui_ScorePanel.cpp
YaLTeR: So it won't build up on disconnect and reconnect etc?

cl_dll/vgui_ScorePanel.h
YaLTeR: Hmm, I suppose, could also be an issue only with one libc implementation.

cl_dll/vgui_ScorePanel.cpp
tmp64: Viewport and ScorePanel are constructed only once (first time connecting to a server) so fonts are only created once.

YaLTeR:
Giving this a test, this font size seems abnormally big compared to what it used to be:

image

tmp64:
WARNING! 8 MB of screenshots ahead.

All VGUI1 fonts are stored in TGAs and can't scale. That means that on large resolutions (1440p, 4K) the scoreboard size scales with the resolutions but fonts stay the same. I set scoreboard text font (used for player rows) to scale proportionally. It makes text more readable but it looks much larger than table header and team names.

I initially used proportional scaling on other fonts but there were some problems with them on low resolutions.

4K screenshots for comparison.
agctf_duff0000

no_remorse0004

I did some testing and I think I've found the best solution.
1. For small resolutions (height <= 768: 640x480, 800x600, 1024x768) it uses font sizes from the scheme (XXX_textscheme.txt). This makes text look nice on low resolutions.
2. For large resolutions (height > 768) font sizes are scaled proportionally but initial font sizes (before scaling) are 2px smaller. This makes it look nice on large resolutions.

1600x900
crossfire0001

1920x1080
crossfire0002

2560x1440
crossfire0003

3840x2160
crossfire0004

PR #86, opened 2020-10-21
Implement ducktap autofunc

mxpph:
Adds a command, `+ducktap` which works as an automatic perfect duckroll. 
Also adds a cvar `cl_autojump_priority`, which controls whether ducktap or jump take precedence.

### DIFF
--- a/cl_dll/input.cpp
+++ b/cl_dll/input.cpp
@@ -66,7 +66,7 @@ cvar_t	*cl_yawspeed;
 cvar_t	*cl_pitchspeed;
 cvar_t	*cl_anglespeedkey;
 cvar_t	*cl_vsmoothing;
-cvar_t	*cl_ducktap_priority;
+cvar_t	*cl_autojump_priority;
 
 namespace autofuncs
 {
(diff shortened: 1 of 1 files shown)


YaLTeR:
> Also adds a cvar cl_ducktap_priority, which controls whether ducktap or jump take precedence.

Is this actually needed by somebody?

mxpph:
> > Also adds a cvar cl_ducktap_priority, which controls whether ducktap or jump take precedence.
> 
> Is this actually needed by somebody?

Begi asked for it so I figured it would be a good idea

cl_dll/input.cpp
```diff
@@ -778,7 +796,12 @@ void CL_DLLEXPORT CL_CreateMove ( float frametime, struct usercmd_s *cmd, int ac
 	//
 	cmd->buttons = CL_ButtonBits( 1 );
 
-	autofuncs::handle_autojump(cmd);
+	if (in_ducktap.state & 1)
+	{
+		cmd->buttons |= IN_DUCK;
+		autofuncs::handle_ducktap(cmd);
+	} else 
+		autofuncs::handle_autojump(cmd);
```
YaLTeR, line 804: Could you add a comment here saying that autojump overrides ducktap?

YaLTeR:
Thanks!

PR #88, opened 2020-10-23
Add unicode server name to the scoreboard

tmp64:
Replaces `vgui::TextImage` with `UnicodeTextImage` in scoreboard server name label. This adds Unicode support, anti-aliasing and high-resolution scaling to the server name shown in the scoreboard.

Before (rendered in 4K, downscaled to 1080p):
crossfire0004

After:
crossfire0007

### DIFF
--- a/cl_dll/vgui_ScorePanel.cpp
+++ b/cl_dll/vgui_ScorePanel.cpp
@@ -137,6 +137,14 @@ ScorePanel::ScorePanel(int x,int y,int wide,int tall) : Panel(x,y,wide,tall)
 	m_TitleLabel.setFgColor( Scheme::sc_primary1 );
 	m_TitleLabel.setContentAlignment( vgui::Label::a_west );
 
+	if (!UnicodeTextImage::shouldFallback())
+	{
+		m_pTitleImage = new UnicodeTextImage();
+		m_pTitleImage->setFont(m_UTitleFont, tfont);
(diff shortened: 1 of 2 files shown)


cl_dll/vgui_ScorePanel.cpp
```diff
@@ -268,6 +276,9 @@ ScorePanel::~ScorePanel()
 #endif
 	if (m_pFlagIcon)
 		delete m_pFlagIcon;
+
+	m_TitleLabel.setImage(nullptr);
+	delete m_pTitleImage;
```
YaLTeR, line 281: Shouldn't this be guarded by shouldFallback?

tmp64: Not really. `vgui::Label` has no image set by default and deleting a null pointer is completely safe.

YaLTeR: I don't see `m_pTitleImage` initialized to `nullptr` anywhere?

tmp64: That's my mistake. I initially was going to always construct UnicodeTextImage and changed the code later. That would actually be a really subtle bug since most of the time `shouldFallback` would return false (and also, ScorePanel is never destroyed).

YaLTeR:
Thanks!

PR #92, opened 2020-11-03
Scoreboard font fixes

tmp64:
1) Fonts on Linux are created 2px taller than they need to be so I added some code to compensate for that.
2) Widened SteamID column by 4px so long SteamIDs don't get clipped.
3) Font names are now loaded from current scheme file (XXX_textscheme.txt) to allow users to change fonts to different ones.

### DIFF
--- a/cl_dll/vgui_SchemeManager.h
+++ b/cl_dll/vgui_SchemeManager.h
@@ -32,6 +32,7 @@ public:
 
 	// getting info from schemes
 	vgui::Font *getFont( SchemeHandle_t schemeHandle );
+	const char *getFontName( SchemeHandle_t schemeHandle );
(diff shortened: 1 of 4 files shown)


cl_dll/vgui_SchemeManager.h
```diff
@@ -1,4 +1,4 @@
-//========= Copyright � 1996-2002, Valve LLC, All rights reserved. ============
```
YaLTeR, line 1: uh

YaLTeR:
Could you squash the last commit into the relevant commits please?

tmp64:
I recreated the commits with fixed tabulation and copyright symbol untouched. The IDE broke it when I saved the file so I replaced it with (c) and accidentally commited the change.

YaLTeR:
Thanks!

PR #94, opened 2020-11-30
Fix incorrect handling of custom view models

### LINKED ISSUE #93: Crash Spectate Mode
The SPECTATE mode does not work correctly, the hl.exe crashes with the servers that use HLBUGFIXED when you are in spectator watching a player with the first person camera.
https://youtu.be/7YY1DZ4wFc0

Example server where crashes IP -> 45.235.98.222:27036

tmp64:
This pull request fixes a few issues with `V_FindViewModelByWeaponModel` function.

1. Fix a crash when trying to find a custom model that is not on a hardcoded list.
2. If model was not found on the list (e.g. a custom WeaponMod weapon), it will attempt to convert player model name to a view model name ("models/p_weapon.mdl" -> "models/v_weapon.mdl"). It works most of the time but some weapons on some server use different names for world and view models (e.g. "p_dbarrel.mdl" and "v_dbarrel_hev.mdl"). In that case spectator won't see the view model.

Resolves #93

### DIFF
--- a/cl_dll/view.cpp
+++ b/cl_dll/view.cpp
@@ -1436,13 +1436,24 @@ int V_FindViewModelByWeaponModel(int weaponindex)
 		int len = strlen( weaponModel->name );
 		int i = 0;
 
-		while ( modelmap[i] != NULL )
+		while ( modelmap[i][0] != NULL )
 		{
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks very much for the fix!

PR #99, opened 2021-03-23
Add savemap command

czarsonszef:
Added savemap command that writes the current map name to a .txt file, mainly intended for HLKZ.

### DIFF
--- a/cl_dll/hud.cpp
+++ b/cl_dll/hud.cpp
@@ -301,7 +301,7 @@ void __CmdFunc_Append()
 	EngineClientCmd(gEngfuncs.Cmd_Argv(1));
 }
 
-void __CmdFunc_Savemap()
+void __CmdFunc_Writemap()
 {
 	FILE* saved_maps;
 	char map_name[64];
@@ -311,7 +311,7 @@ void __CmdFunc_Savemap()
 
 	if (map_name[0])
 	{
-		saved_maps = fopen("saved_maps.txt", "r+"); // returns NULL if the file doesn't exist or if the file is read-only
+		saved_maps = fopen("saved_maps.txt", "r+");
 
 		if (saved_maps)
 		{
@@ -467,7 +467,7 @@ void CHud :: Init( void )
 
 	HOOK_COMMAND( "agrecord", Agrecord );
 	HOOK_COMMAND( "append", Append );
-	HOOK_COMMAND( "savemap", Savemap );
+	HOOK_COMMAND( "writemap", Writemap );
 	EngineClientCmd("alias zpecial \"append _zpecial\"");
 
 	force_model::hook_commands();
--- a/cl_dll/hud.cpp
+++ b/cl_dll/hud.cpp
@@ -301,6 +301,38 @@ void __CmdFunc_Append()
 	EngineClientCmd(gEngfuncs.Cmd_Argv(1));
 }
 
+void __CmdFunc_Savemap()
+{
+	FILE* saved_maps;
+	char map_name[64];
+	char map_name_to_check[64];
+
+	get_map_name(map_name, ARRAYSIZE(map_name));
+	sprintf(map_name, "%s\n", map_name);
+
+	if (map_name[0])
+	{
+		saved_maps = fopen("saved_maps.txt", "r");
+
+		if (saved_maps)
+		{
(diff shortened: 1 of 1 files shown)


YaLTeR:
Don't mind the CI, looks unrelated.

What's the use-case for this?

czarsonszef:
While playing HLKZ i often found myself saving map names in notepad to play them later, thought it might be useful for kz players.

cl_dll/hud.cpp
```diff
@@ -301,6 +301,38 @@ void __CmdFunc_Append()
...
+{
+	FILE* saved_maps;
+	char map_name[64];
+	char map_name_to_check[64];
+
+	get_map_name(map_name, ARRAYSIZE(map_name));
+
+	if (map_name[0])
+	{
+		sprintf(map_name, "%s\n", map_name);
```
YaLTeR, line 314: If the map is 63 characters long + null byte, this will write one byte past the end of the array. I suggest comparing without `\n` below rather than appending `\n` here.

YaLTeR (changes requested):

cl_dll/hud.cpp
```diff
@@ -301,6 +301,38 @@ void __CmdFunc_Append()
...
+				if (!strcmp(map_name, map_name_to_check))
+				{
+					gEngfuncs.Con_Printf("Current map is already in saved_maps.txt");
+					fclose(saved_maps);
+					return;
+				}
+			}
+		}
+
+		saved_maps = fopen("saved_maps.txt", "a");
```
tmp64, line 330: Wouldn't this leak the file opened by previous `fopen` call?

YaLTeR: Good point. Also when opening the file for a second time, the result needs to be checked too.

I suggest not opening the file twice, but opening it once for read+write and then seeking to the end to write.

czarsonszef:
I've made the file open in read/write mode as requested. If the file doesn't exist it gets created by calling fopen() in append mode, if the file didn't exist before it doesn't need to be checked for the presence of the map name.

cl_dll/hud.cpp
czarsonszef: > Wouldn't this leak the file opened by previous `fopen` call?

If im correct theres now no possibility of such leak occuring, because fclose() doesn't need to be called after opening a file in r+ mode that doesn't exist.

cl_dll/hud.cpp
```diff
@@ -301,6 +301,42 @@ void __CmdFunc_Append()
...
+				{
+					gEngfuncs.Con_Printf("Current map is already in saved_maps.txt\n");
+					fclose(saved_maps);
+					return;
+				}
+			}
+		}
+		else
+		{
+			saved_maps = fopen("saved_maps.txt", "a"); // creates the file if it doesn't exist
```
YaLTeR, line 332: I'd put a check here too. For example if the player accidentally made their folder read-only I imagine this could fail.

cl_dll/hud.cpp
```diff
@@ -301,6 +301,42 @@ void __CmdFunc_Append()
...
+
+				if (!strcmp(map_name, map_name_to_check))
+				{
+					gEngfuncs.Con_Printf("Current map is already in saved_maps.txt\n");
+					fclose(saved_maps);
+					return;
+				}
+			}
+		}
+		else if(!(saved_maps = fopen("saved_maps.txt", "a"))) // checks if the folder is read-only. If yes - saved_maps is NULL, if not - creates the file
```
YaLTeR, line 330: Uhhh, sorry for nitpicking but could you please not use assignment inside an if condition? It frequently leads to hard to notice bugs. Other than that this PR is fine.

```c++
else
{
    saved_maps = fopen("saved_maps.txt", "a");
    if (!saved_maps)
        return;
}
```
should be sufficient.

execut4ble:
Just a suggestion, perhaps call it "writemap" so that it would be consistent with other commands such as writecfg, writeid, writeip?

czarsonszef:
Do you think it would be more suitable to write the map name to a .cfg file?

YaLTeR:
> Do you think it would be more suitable to write the map name to a .cfg file?

No, I think .txt is perfectly fine.

YaLTeR (approved):

YaLTeR:
Thanks!

Issue #90, opened 2020-10-30
hud_color rainbow


SmileyAG:
A cvar to make hud rainbow.

LeonN1:
Having something like this would be amazing
![IMG-20210115-WA0027](https://user-images.githubusercontent.com/77511070/104775191-f46bd680-5745-11eb-92d9-e3f6a1f07ae0.jpeg)

SmileyAG:
Sorry for too late response, i'll closing this issue, since tmp64 implemented this feature already, big thanks to him!

PR #100, opened 2021-04-01
Add Rainbow HUD

tmp64:
Adds a number of cvars:
- `hud_rainbow`: enables rainbow HUD (default on)
- `hud_rainbow_sat`: saturation of rainbow HUD (default 100%)
- `hud_rainbow_val`: brightness of rainbow HUD (default 100%)
- `hud_rainbow_speed`: angular speed of rainbow HUD (default: 40 degrees/s)
- `hud_rainbow_xphase`: horizontal phase shift of rainbow HUD (default: 0.4 degrees/px)
- `hud_rainbow_yphase`: vertical phase shift of rainbow HUD (default: 0.7 degrees/px)

**Demonstration:**
*Note: Rainbow GameUI not included*
https://youtu.be/i_CHv75CZC4

That's what issue #90 was about, right?

### DIFF
--- a/cl_dll/ammohistory.cpp
+++ b/cl_dll/ammohistory.cpp
@@ -125,12 +125,16 @@ int HistoryResource :: DrawAmmoHistory( float flTime )
 
 				int r, g, b;
 				UnpackRGB(r,g,b, gHUD.m_iDefaultHUDColor);
-				float scale = (rgAmmoHistory[i].DisplayTime - flTime) * 80;
-				ScaleColors(r, g, b, min(scale, 255) );
 
 				// Draw the pic
 				int ypos = ScreenHeight - (AMMO_PICKUP_PICK_HEIGHT + (AMMO_PICKUP_GAP * i));
 				int xpos = ScreenWidth - 24;
+
+				if (gHUD.IsRainbow())
+					gHUD.GetRainbowColor(xpos, r, g, b);
+				float scale = (rgAmmoHistory[i].DisplayTime - flTime) * 80;
+				ScaleColors(r, g, b, min(scale, 255));
+
 				if ( spr && *spr )    // weapon isn't loaded yet so just don't draw the pic
 				{ // the dll has to make sure it has sent info the weapons you need
 					SPR_Set( *spr, r, g, b );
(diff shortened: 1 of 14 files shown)


cl_dll/ammohistory.cpp
```diff
@@ -125,19 +125,25 @@ int HistoryResource :: DrawAmmoHistory( float flTime )
...
+
 				if ( spr && *spr )    // weapon isn't loaded yet so just don't draw the pic
 				{ // the dll has to make sure it has sent info the weapons you need
 					SPR_Set( *spr, r, g, b );
 					SPR_DrawAdditive( 0, xpos, ypos, &rcPic );
 				}
 
 				// Draw the number
+				if (gHUD.IsRainbow())
+					gHUD.GetRainbowColor(xpos, r, g, b);
```
YaLTeR, line 146: xpos - 10?

YaLTeR (commented):
lol that's actually great

tmp64:
Rewrote the whole thing. Instead of modifying all HUD elements to use rainbow color, I point sprite and text rendering functions in `gEngfuncs` to custom wrappers that change the colors. This makes it easier to maintain the code. The only place that uses `GetRainbowColor` is CHudMessage since it adds blending to original colors for effects.

Also added `hud_rainbow_yphase` cvar for vertical phase shift and saturation and value clamping so nothing breaks on invalid input.

cl_dll/rainbow.cpp
```diff
@@ -0,0 +1,216 @@
...
+    // Register cvars
+    m_pCvarRainbow = CVAR_CREATE("hud_rainbow", "0", FCVAR_ARCHIVE);
+    m_pCvarRainbowS = CVAR_CREATE("hud_rainbow_sat", "100", FCVAR_ARCHIVE);
+    m_pCvarRainbowV = CVAR_CREATE("hud_rainbow_val", "100", FCVAR_ARCHIVE);
+    m_pCvarRainbowSpeed = CVAR_CREATE("hud_rainbow_speed", "40", FCVAR_ARCHIVE);
+    m_pCvarRainbowXPhase = CVAR_CREATE("hud_rainbow_xphase", "0.4", FCVAR_ARCHIVE);
+    m_pCvarRainbowYPhase = CVAR_CREATE("hud_rainbow_yphase", "0.7", FCVAR_ARCHIVE);
+
+    // Save engine functions
+    m_pfnSPR_Set = gEngfuncs.pfnSPR_Set;
```
YaLTeR, line 20: Is it in any way possible for this to be called while functions are hooked? E.g. `_restart`. Because then it'll break.

tmp64: No, it's not. Engine unloads the client when shutting down. And if unloading fails, then `Initialize`, which initializes `gEngfuncs`, is always called before `Init`.

PR #101, opened 2021-04-03
Save default_fov into the config

tmp64:
Flags `default_fov` cvar to be written into the game config so players don't have to hardcode it in the user config.


YaLTeR:
Thanks

PR #104, opened 2021-04-17
Fixed spectator mode VGUI panel

chinese-soup:
Hello. Useless fixes galore!

This PR fixes a thing that's probably only been bugging me as I am probably the only person to ever try to use the VGUI spectator panel in AG and it always ticked me off that it doesn't work. This is the panel that appears when you press +duck while in spectator mode.

Let's get this out of the way first: spectating while on a server (and not on an HLTV server or on dem_forcehltv 1) is **server-side**, the **client** only sends the server commands when it wants to change the target to spectate / change the mode of spectating etc.

### First: let's start with the easy stuff. To fix this menu:
image
This change has to be made:
```			
sprintf(cmdstring,"spec_mode %i",iNewMainMode );
```
As the server DLL that we are all using implements this client->server cmd as "spec_mode" instead of "specmode", see here: https://github.com/martinwebrant/agmod/blob/master/src/dlls/agclient.cpp#L126

### Second: these two arrows were already working, but didn't make sense, they were flipped, i.e. the NEXT did previous player and PREV did next player, I fixed that in VGUI_SpectatorPanel.cpp**
image

### Third of all: the worst offender - the player picker ###
image

This panel is supposed to just send the name of the player to `FindPlayer(name)` that sends the `follow %s` command to the AG server dll, but the AG server dll removed this and **doesn't implement this at all**, it only implements the `follownext %i` command (that's why clicking with the +attack & +attack2 works and also why the arrows mentioned above work) as can be seen here:
https://github.com/martinwebrant/agmod/blob/master/src/dlls/agclient.cpp#L120

**The workaround** for this is to calculate how many times the client needs to send the `follownext %i` command where %i decides whether to go backwards (previous player) (1) or forwards (next player) (0).

**Quirks & stuff:**
* The hell with this was that the `g_iUser2` variable doesn't get updated as it's only updated once the server messages the client about the change and we don't get that until after the FindPlayer function already returns. This is the reason why we can't just do `while(g_iUser2 != target_player_id)` as we'd just get stuck in an infinite loop.

* Another hell is that we don't know what players are active (i.e. are playing and not spectating/are on the server/etc.) until I go through all of them, hence the requirement of the first `for loop`.

* I couldn't think of a better way to implement this than having two seperate vectors, for now anyway, if it's requested I will try to rewrite. 
* The main reason for the two vectors is that it was easier to me than "saving an index, iterating forwards, and the player wasn't found, iterating backwards again" in on a single vector. 
* You can't just do a difference between the g_iUser2 and the target player id as the server can have **player ids that are not consecutive** (e.g. players 1,2,6,9), not to mention that **not all connected players are spectatable** (i.e. they are also spectating).
* Or, I do realize I could just overflow(? dunno if it's the right word -- basically start the index at <currently spectated player id> and if I get to the end just reset the index from the beginning and try find the player there) the number of times the `follownext 0` (just going forwards) is called until I get to the players that are behind the one I'm currently spectating, but I just couldn't get it to work that way for way too long, so I literally gave up and did this stupidity with forwards/backwards instead. 
* But like I've said, if a rewrite is requested, I'm willing to try it again. :smile_cat: 

* This has been tested with player IDs that are not consecutive (e.g. 1,2,3,5) which can be seen in the example video below where the player with the id = 4 has been kicked. 
* This also works if the user has colors in the name, as the VGUI SpectatorPanel doesn't **show** the colors in the combobox,  but its action handler **sends the original nickname including colors.** 
* This also works if a player we're spectating is kicked/quit/changed to spectator, as that's handled elsewhere and makes us spectate the next available player (or switches to Free look etc.)
* This doesn't break +attack or +attack2 as those just call **hud_spectator.cpp:FindNextPlayer()**, not this function.

All the fixes can be seen featured in this video: **https://user-images.githubusercontent.com/5108747/115096852-5562a080-9f27-11eb-8357-d83127154249.mp4**

Review welcome, I am a big C++ (or well, HLSDK is mostly just C with OOP but whatever) noob and this is a learning experience for me. 

PS: The player picker could obviously be fixed server side by adding back the "follow" command that the AGMod author deleted. But for now, here's this.

### DIFF
--- a/cl_dll/hud_spectator.cpp
+++ b/cl_dll/hud_spectator.cpp
@@ -990,10 +990,82 @@ void CHudSpectator::FindPlayer(const char *name)
 	// if we are NOT in HLTV mode, spectator targets are set on server
 	if ( !gEngfuncs.IsSpectateOnly() )
 	{
+		/*
+		 * HACK HACK HACK: AG server DLL doesn't let us "follow" a player with a specific name, but only lets us
+		 * use "follownext". This code uses follownext several times until it gets to the specific player
+		 * that the user clicked in the VGUI menu.
+		 *
+		 * Cannot use while(g_iUser2.name == current.name) because g_iUser doesn't get updated fast enough
+		 * (it's a server msg after all) which results in an infinite loop.
+		 *
+		 * If the user picks a player that has an ID that's higher than the one we're currently spectating (g_iUser2)
+		 * we step forward (reverse = 0)
+		 * If the user picks a player that has an ID that's lower than the one we're currently spectatng (g_iUser2)
+		 * we step backwards (reverse = 1)
+		 *
+		 */
+
+		std::vector<std::string> player_names;
+		std::vector<std::string> player_names_backwards;
+
+		cl_entity_t * pEnt = NULL;
+		gViewPort->GetAllPlayersInfo();
+		for (int i = 1; i < MAX_PLAYERS; i++ )
+		{
+			pEnt = gEngfuncs.GetEntityByIndex(i);
+
+			if (!gHUD.m_Spectator.IsActivePlayer(pEnt))
+				continue;
+
+			// Our user is already spectating this player, let's bail out
+			if (!strcmp(g_PlayerInfoList[i].name, name) && i == g_iUser2)
+				return;
+
+			// The current (i) player's id is higher than the one the user is currently spectating, let's save him to the forward gang
+			if ( i > g_iUser2 )
+				player_names.push_back(g_PlayerInfoList[i].name);
+
+			// The current (i) player's id is lower than the one the user is currently spectating, let's save him to the backward gang
+			else if ( i < g_iUser2 )
+				player_names_backwards.insert(player_names_backwards.begin(), g_PlayerInfoList[i].name);
+		}
+
+		int reverse = -1; // -1 so we can tell if we found the player going forwards or backwards
+		int steps = 1; // We start at 1 because the player we are currently spectating is NOT present in the list
+
+		for ( auto &player_name : player_names )
+		{
+			if(player_name == name)
+			{
+				reverse = 0; // set reverse to false, we're gonna go forward to get to our target player
+				break;
+			}
+			else
+				steps += 1;
+		}
+
+		if(reverse == -1)
+		{
+			steps = 1; // reset the steps we got from forwards
+			for ( auto &player_name_backwards : player_names_backwards )
+			{
+				if(player_name_backwards == name)
+				{
+					reverse = 1; // set reverse to true, we're gonna go backwards to get to our target player
+					break;
+				}
+				else
+					steps += 1;
+			}
+		}
+
 		char cmdstring[32];
-		// forward command to server
-		sprintf(cmdstring,"follow %s",name);
-		gEngfuncs.pfnServerCmd(cmdstring);
+		for(int i=0; i < steps; i++)
+		{
+			// forward command to server
+			sprintf(cmdstring,"follownext %i", reverse);
+			gEngfuncs.pfnServerCmd(cmdstring);
+		}
 		return;
 	}
 	
@@ -1182,7 +1254,10 @@ void CHudSpectator::SetModes(int iNewMainMode, int iNewInsetMode)
 		{
 			char cmdstring[32];
 			// forward command to server
-			sprintf(cmdstring,"specmode %i",iNewMainMode );
+			if (!strcmp(gEngfuncs.pfnGetGameDirectory(), "ag"))
+				sprintf(cmdstring,"spec_mode %i",iNewMainMode );
+			else
+				sprintf(cmdstring,"specmode %i",iNewMainMode );
 			gEngfuncs.pfnServerCmd(cmdstring);
 			return;
 		}
--- a/cl_dll/vgui_SpectatorPanel.cpp
+++ b/cl_dll/vgui_SpectatorPanel.cpp
@@ -61,10 +61,10 @@ void SpectatorPanel::ActionSignal(int cmd)
 		case SPECTATOR_PANEL_CMD_OPTIONS :		gViewPort->ShowCommandMenu( gViewPort->m_SpectatorOptionsMenu );
 												break;
 
-		case SPECTATOR_PANEL_CMD_NEXTPLAYER :	gHUD.m_Spectator.FindNextPlayer(true);
+		case SPECTATOR_PANEL_CMD_NEXTPLAYER :	gHUD.m_Spectator.FindNextPlayer(false);
 												break;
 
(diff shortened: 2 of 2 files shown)


cl_dll/hud_spectator.cpp
```diff
@@ -990,10 +990,82 @@ void CHudSpectator::FindPlayer(const char *name)
...
+		}
+
 		char cmdstring[32];
-		// forward command to server
-		sprintf(cmdstring,"follow %s",name);
-		gEngfuncs.pfnServerCmd(cmdstring);
+		for(int i=0; i < steps; i++)
+		{
+			// forward command to server
+			sprintf(cmdstring,"follownext %i", reverse);
```
YaLTeR, line 1066: I wonder if you could chain server cmds together with a `;` like you can do clientside. If not, is it possible that this can kick you for flooding or something?

cl_dll/hud_spectator.cpp
```diff
@@ -990,10 +990,82 @@ void CHudSpectator::FindPlayer(const char *name)
 	// if we are NOT in HLTV mode, spectator targets are set on server
 	if ( !gEngfuncs.IsSpectateOnly() )
 	{
+		/*
+		 * HACK HACK HACK: AG server DLL doesn't let us "follow" a player with a specific name, but only let's us
```
YaLTeR, line 994:
```suggestion
		 * HACK HACK HACK: AG server DLL doesn't let us "follow" a player with a specific name, but only lets us
```

YaLTeR (commented):
The `spec_mode` change and the next/prev arrow change could be 2 separate commits btw if it's not too much of a bother.

> that the AGMod author deleted

Might just be that this wasn't a thing in the old HLSDK when AG was made. For the repo Martin just rebased his work on top of the latest HLSDK, so it's very likely that he didn't go through all the new features to adjust the AG code accordingly.

YaLTeR:
@tmp64 would you mind reviewing this too?

cl_dll/hud_spectator.cpp
tmp64: I just tested: you can't. The server will think that `test1;test2;test3` is one long command and not three separate ones. But I don't know if it can cause the client to be kicked.

cl_dll/hud_spectator.cpp
```diff
@@ -1182,7 +1254,7 @@ void CHudSpectator::SetModes(int iNewMainMode, int iNewInsetMode)
 		{
 			char cmdstring[32];
 			// forward command to server
-			sprintf(cmdstring,"specmode %i",iNewMainMode );
+			sprintf(cmdstring,"spec_mode %i",iNewMainMode );
```
tmp64, line 1257: This will break SetModes for players that use OpenAG DLL on a Half-Life client. I would add a gamedir check to use `spec_mode` in AG and `specmode` in HL and other mods.
```cpp
if (!strcmp(gEngfuncs.pfnGetGameDirectory(), "ag"))
  sprintf(cmdstring,"spec_mode %i",iNewMainMode );
else
  sprintf(cmdstring,"specmode %i",iNewMainMode );
```

cl_dll/hud_spectator.cpp
chinese-soup: Hello. If you mean checks for flooding that are implemented in the AG server dll, then I don't think that would happen, this command isn't checked for flooding as opposed to e.g. "spectate", the difference can can be seen here:
https://github.com/martinwebrant/agmod/blob/master/src/dlls/agclient.cpp#L102-L125

If you meant a server code unrelated to AG (just in HL in general), I don't think they implement an anti-flood like that.
Just did a quick test with:
```
		for(int i=0; i < 5000; i++)
		{
				_snprintf( mrdka, sizeof( mrdka ), "i = %f\n", gEngfuncs.GetClientTime());
				gEngfuncs.pfnConsolePrint( mrdka );
				// forward command to server
				sprintf(cmdstring,"follownext %i", reverse);
				gEngfuncs.pfnServerCmd(cmdstring);
		}
```
Which is more than it's ever gonna call, seeing as there's only 32 players max. The client is still on the server (no kick or anything), the change gets noticed by the server, I think it drops the last few ones? But it doesn't kick me (local HLDS), the only thing that points to something being "wrong" is the HLDS log that says:
```
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
NET_QueuePacket:  Oversize packet from 192.168.0.59:27005
```
And the client with developer 1 says:
```
Netchan_Transmit:  Unreliable would overfow, ignoring
```
Which I'd assume means that it never hits the server with the ones that are already overflowed? You probably know better, but anyway, I think the chaining doesn't matter as it looks like the client (and server also?) just buffer the commands themselves. :)
 
Anyhow, this shouldn't be a problem for AG as maxplayers isn't 5000, but 32? Tested with a sane 32 and I end up spectating the same person correctly and everything all 32 commands are sent (I assume so, 
```
		for(int i=0; i < 32; i++)
		{
				_snprintf( mrdka, sizeof( mrdka ), "i = %f\n", gEngfuncs.GetClientTime());
				gEngfuncs.pfnConsolePrint( mrdka );
				// forward command to server
				sprintf(cmdstring,"follownext %i", i % 2);
				gEngfuncs.pfnServerCmd(cmdstring);
		}
```
I added that ^ and clicked the fastest way I could in the panel and it seemed to send it all and accept it server side. I also tried it on a random server with 4 players playing and the server was fine having me there. :)

I'd say we should be fine, although my test isn't very scientific.

cl_dll/hud_spectator.cpp
chinese-soup: Alright, then this check has to also be added for the `FindPlayer` function, where I replaced "follow" with the "follownext" stuff.

chinese-soup: Although, specifically, for the "spec mode" command wouldn't it be easier to send both specmode and spec_mode and let the server decide the "correct answer"?

tmp64: I thought about it but it will spam "Unknown command" messages in the console

chinese-soup: Which it already did to OpenAG client players on AG servers. :smile: Anyway, ok, will check for the mod dir each time, thanks.

YaLTeR:
This should be good to merge now, right?

chinese-soup:
Not yet, sorry.

chinese-soup:
Never mind, it is. Turns out the "follow" command isn't implemented in vanilla Half-Life server DLL either, so there's no need to check for gamemode there. This is good to go now.

YaLTeR:
Alright, cool, thanks!

Issue #89, opened 2020-10-30
cl_old_scoreboard 1


SmileyAG:
A cvar for a redesign scoreboard like in AG 6.6 (implemenation can be get here: https://github.com/martinwebrant/agmod/blob/master/src/cl_dll/aghudscoreboard.cpp)
![1](https://user-images.githubusercontent.com/58108407/97696166-9c1a2680-1abe-11eb-8c7f-6fd038536bd5.jpg)
![2](https://user-images.githubusercontent.com/58108407/97696170-9cb2bd00-1abe-11eb-878b-d8beef84df91.PNG)

PR #105, opened 2021-04-17
Fixed HUD align bug in ammo history for ammo < 10 (#54)

### LINKED ISSUE #54: Aligning bug in HUD displaying ammo if amount is < 10
pictures talk for themselves

chinese-soup:
Fixes #54 

**Screenshots:**
image
image

### DIFF
--- a/cl_dll/ammohistory.cpp
+++ b/cl_dll/ammohistory.cpp
@@ -138,7 +138,9 @@ int HistoryResource :: DrawAmmoHistory( float flTime )
 				}
 
 				// Draw the number
-				gHUD.DrawHudNumberString( xpos - 10, ypos, xpos - 100, rgAmmoHistory[i].iCount, r, g, b );
+				char szString[32];
+				snprintf( szString, ARRAYSIZE(szString), "%d", rgAmmoHistory[i].iCount );
+				gHUD.DrawHudStringRightAligned( xpos - 10, ypos, szString, r, g, b );
 			}
 			else if ( rgAmmoHistory[i].type == HISTSLOT_WEAP )
 			{


cl_dll/ammohistory.cpp
```diff
@@ -138,7 +138,9 @@ int HistoryResource :: DrawAmmoHistory( float flTime )
 				}
 
 				// Draw the number
-				gHUD.DrawHudNumberString( xpos - 10, ypos, xpos - 100, rgAmmoHistory[i].iCount, r, g, b );
+				char szString[32];
+				sprintf( szString, "%d", rgAmmoHistory[i].iCount );
```
YaLTeR, line 142: Just in case could you use `snprintf` here please?

YaLTeR (changes requested):

cl_dll/ammohistory.cpp
chinese-soup: For sure, it was a direct copy paste from DrawHudNumberString. Done.

chinese-soup: @YaLTeR I've noticed that the `DrawHudNumberString` function is no longer called by anything as it was only called in this instance, should I change it there instead?

YaLTeR: Yeah, that would make sense I guess.

chinese-soup: Never mind, it's used in `scoreboard.cpp` which isn't used _right now_, but could be if one were to implement this: https://github.com/YaLTeR/OpenAG/issues/89 so I'd say for compatibility, let's keep the DrawHudNumberString unchanged. Sorry about that.

YaLTeR: So this PR is good to go now right?

chinese-soup: Yep.

YaLTeR:
Thanks!

Issue #55, opened 2018-02-18
cl_forceteammatemodel doesnt work in spectate mode


krangm:
when using `cl_forceenemymodel` and `cl_forceteammatemodel` and viewing from spectate, all players in any team have models set in `cl_forceenemymodel`

YaLTeR:
That's because when you're spectating you aren't in any team (you are in a separate "Spectators" team).

krangm:
supposed so
but is it fixable?

YaLTeR:
Which team would it use instead?

krangm:
team named same as your player model

krangm:
bump

@YaLTeR make it please

YaLTeR:
I'll make it whenever I have time and feel like working on OpenAG next time. In the meantime you're welcome to make it yourself and pull request. :slightly_smiling_face:

krangm:
@YaLTeR OpenAG update when it ?

chinese-soup:
Implemented by #106

PR #106, opened 2021-04-17
Forcemodel: Treat players with our model as our teammates when spectating

chinese-soup:
Hello. This implements #55 in a very, very crude way. You decide whether this is worth it.

The motivation behind this can be read in this copypasted code comment:

Hack.

We cannot get the correct model from `g_PlayerInfoList[local_player_index].model`
as it always shows the one we had BEFORE we changed to a new team.

I.e. if one would use `g_PlayerInfoList[local_player_index].model` here, it would only work with the team
 you were in BEFORE you switched to spectator, and any changes of the team while being in spectator
won't propagate soon enough into` g_PlayerInfoList[local_player_index].model` so we can check it here.
No amount of careful `gEngfuncs.pfnGetPlayerInfo( clientIndex, &g_PlayerInfoList[clientIndex] );`
placements unfortunately helps as per my testing.

`gEngfuncs.GetLocalPlayer()->model` can't be used either as that just always has models/player.mdl.

E.g. 
1) You are playing on team "blue" with a teammate against a team red.

2) You switch to spectator

3) You are on **model** `blue` according to `g_PlayerInfoList[local_player_index].model`
You are on **team** `(null)` according to `g_PlayerInfoList[local_player_index].teamname`

4) You switch to team **red** **while spectating (**`model red `in console)

5) You **STILL** are on model **blue** according to `g_PlayerInfoList[local_player_index].model `at the time of this function being called.
			

**Console log showing the behavior:**
```
* you are on team 'yep'
Unknown command: specmode
] spectate
Your team name is = 
Your model name is = yep
Your team name is = 
Your model name is = yep
asdf entered spectator mode
UserMsg: No pfn Splash 103
[HLKZ] Welcome to HL KreedZ Beta. MAP: cg_cbblebhop
[HLKZ] Visit the pooping.men Public Mumble Server
[HLKZ] You can say /kzhelp to see available commands
] model blue
Your team name is = 
Your model name is = yep
Your team name is = 
Your model name is = yep
```

* In this log you can see that I join the server on team "yep", I then go spectate, my model in `g_PlayerInfoList[local_player_index].model` is printed out when i switch to spectate team ((null)) correctly and could be used for setting the teammate models. 
* However as soon as I change the model WHILE still spectating you can see it clearly prints out the "yep" team again. If I were to change teams again at this point to for example "red" the debug print would print out "Your model name is = blue" aka the one before the change again.

### DIFF
--- a/cl_dll/forcemodel.cpp
+++ b/cl_dll/forcemodel.cpp
@@ -477,6 +477,40 @@ namespace force_model
 		}
 
 		const auto local_player_index = gEngfuncs.GetLocalPlayer()->index;
+		if (g_IsSpectator[local_player_index])
+		{
+			/*
+			 * Hack.
+			 *
+			 * We cannot get the correct model from g_PlayerInfoList[local_player_index].model
+			 * as it always shows the one we had BEFORE we changed to a new team.
+			 *
+			 * I.e. if one would use g_PlayerInfoList[local_player_index].model here, it would only work with the team
+			 * you were in BEFORE you switched to spectator, and any changes of the team while being in spectator
+			 * won't propagate soon enough into g_PlayerInfoList[local_player_index].model so we can check it here.
+			 * No amount of careful gEngfuncs.pfnGetPlayerInfo( clientIndex, &g_PlayerInfoList[clientIndex] );
+			 * placements unfortunately helps as per my testing.
+			 *
+			 * gEngfuncs.GetLocalPlayer()->model can't be used either as that just always has models/player.mdl.
+			 *
+			 * E.g. 1) You are playing on team "blue" with a teammate against a team red.
+			 * 2) You switch to spectator
+			 * 3) You are on model blue according to g_PlayerInfoList[local_player_index].model
+			 * You are on team (null) according to g_PlayerInfoList[local_player_index].teamname
+			 * 4) You switch to team red while spectating.
+			 * 5) You STILL are on model blue according to g_PlayerInfoList[local_player_index].model at the time of
+			 * this function being called.
+			 */
+			std::string model_name(CVAR_GET_STRING("model"));
+			std::transform(model_name.begin(), model_name.end(), model_name.begin(), ::tolower);
+
+			if (!strcmp(model_name.c_str(), g_PlayerExtraInfo[player_index + 1].teamname))
+				teammate_enemy_model_overrides_cache[player_index] = teammate_model_override;
+			else
+				teammate_enemy_model_overrides_cache[player_index] = enemy_model_override;
+			return;
+		}
+
 		if (!strcmp(g_PlayerExtraInfo[local_player_index].teamname, g_PlayerExtraInfo[player_index + 1].teamname))
 			teammate_enemy_model_overrides_cache[player_index] = teammate_model_override;
 		else


cl_dll/forcemodel.cpp
```diff
@@ -479,6 +479,38 @@ namespace force_model
 		const auto local_player_index = gEngfuncs.GetLocalPlayer()->index;
 		if (!strcmp(g_PlayerExtraInfo[local_player_index].teamname, g_PlayerExtraInfo[player_index + 1].teamname))
 			teammate_enemy_model_overrides_cache[player_index] = teammate_model_override;
+		else if(g_IsSpectator[local_player_index])
```
YaLTeR, line 482: Maybe put this if separately right before? So it's like
```cpp
const auto local_player_index = gEngfuncs.GetLocalPlayer()->index;

if (g_IsSpectator[local_player_index])
{
    // ...
    return;
}

if (!strcmp(g_PlayerExtraInfo[local_player_index].teamname, g_PlayerExtraInfo[player_index + 1].teamname))
```

cl_dll/forcemodel.cpp
```diff
@@ -479,6 +479,38 @@ namespace force_model
...
+			 * E.g. 1) You are playing on team "blue" with a teammate against a team red.
+			 * 2) You switch to spectator
+			 * 3) You are on model blue according to g_PlayerInfoList[local_player_index].model
+			 * You are on team (null) according to g_PlayerInfoList[local_player_index].teamname
+			 * 4) You switch to team red while spectating.
+			 * 5) You STILL are on model blue according to g_PlayerInfoList[local_player_index].model at the time of
+			 * this function being called.
+			 */
+			std::string model_name(CVAR_GET_STRING("model"));
+			transform(model_name.begin(), model_name.end(), model_name.begin(), ::tolower);
```
YaLTeR, line 507: Is there a `using namespace std;` somewhere? Either way:
```suggestion
			std::transform(model_name.begin(), model_name.end(), model_name.begin(), ::tolower);
```

cl_dll/forcemodel.cpp
chinese-soup: Done &

cl_dll/forcemodel.cpp
chinese-soup: & done.

cl_dll/forcemodel.cpp
```diff
@@ -477,6 +477,39 @@ namespace force_model
...
+			 * this function being called.
+			 */
+			std::string model_name(CVAR_GET_STRING("model"));
+			std::transform(model_name.begin(), model_name.end(), model_name.begin(), ::tolower);
+
+			if (!strcmp(model_name.c_str(), g_PlayerExtraInfo[player_index + 1].teamname))
+				teammate_enemy_model_overrides_cache[player_index] = teammate_model_override;
+			else
+				teammate_enemy_model_overrides_cache[player_index] = enemy_model_override;
+		}
```
YaLTeR, line 511: Missing `return;` here.

chinese-soup: Thanks, fixed.

YaLTeR:
Thanks!

PR #108, opened 2021-04-20
Fix viewmodel FOV when weapon is zoomed in

tmp64:


YaLTeR:
Thanks

chinese-soup:
And a thanks from me as well!

PR #109, opened 2021-04-20
Hack fix shotgun shell popping into view when cl_viewmodel variables are used

chinese-soup:
This is some really hacky bullcrap, but I don't have a better idea for this one right now.  We can always revert this if we find a non-hacky way to fix this. Until then I offer this.

**Video** of what happens in the game without the fix:
**[https://user-images.githubusercontent.com/5108747/115474453-bd71f900-a23d-11eb-807d-2de2782af506.mp4](https://user-images.githubusercontent.com/5108747/115474453-bd71f900-a23d-11eb-807d-2de2782af506.mp4)**

**Screenshot** what _basically_ happens in the game without the fix:
image

Copy pasted comment:
```
			// Hack: The shotgun has a shell that gordon reloads in the reload animation as a bodypart
			// This bodypart is visible when projected seperately from the default_fov projection.
			// This unfortunately creates a problem where the shell is directly in front of the view
			// when walking forward/backwards as it's being projected
			// "in front of us", hence when the user uses cl_viewmodel_forward > 0 or any FOV
			// the shell just appears on the screen.
			// This gets fixed by this hack, where the shell is not rendered at all unless
			// the user is actually reloading (reload sequence (3 in v_shotgun.mdl) is the current sequence).
			//
			// IEngineStudio API (nor manually checking the m_pStudioHeader) unfortunately doesn't give us the names
			// of the body parts, API doesn't have anything stored for "name", manually it only returns "studio"
			// and not the name like "shell", the shotgun shell is the second submodel tho, hence the i == 1
```


chinese-soup:
Drafting as im going to bed and don't want to merge as this currently breaks for custom (some) models and HD models (because of the different index of submodels), they are usually based on either the original (so the shell is there) or the HD one (where the shell is in a good position that you wouldn't be able to see ingame aka HD shotgun model don't suffer from this problem).

Anyway, e.g. the official HL HD model for the shotgun has a bodypart with index 1 where that bodyport isn't the shell, but the hands that hold the gun. (And like i've said this shell bug doesn't appear in the HD models).

Anyway point is, i need to get the correct name somehow. dunno how, it just keeps sending "studio" (and like ive said in the comment, m_pBodyPart's name is just random memory.

("lmao" being manual, API being "yass")
```c++
//int CStudioModelRenderer::StudioDrawModel( int flags )

    mstudiobodyparts_t *pbodyparts = (mstudiobodyparts_t *) ((byte *) m_pStudioHeader + m_pStudioHeader->bodypartindex);
        gEngfuncs.Con_Printf( "lmao: %s %s\n", pbodyparts[0].name, pbodyparts->name );
```
```c++
//StudioRenderFinal_Hardware
if (m_pCurrentEntity == gEngfuncs.GetViewModel() && m_pCvarViewmodelFov->value != 0.0f)
			{
			    auto tvojemama = (mstudiobodyparts_t)*m_pBodyPart;
                gEngfuncs.Con_Printf("yass: %d | %s | %s \n", m_pStudioHeader->bodypartindex, &m_pBodyPart->name);
				SetViewmodelFovProjection();

			}
```


```
lmao: parts parts
lmao: parts parts
lmao: parts parts
lmao: studio studio
lmao: studio studio
yass: 30524 | �N�� | �<�����
                            Í�& 
yass: 30524 | O�� | �<�����
                           Í�& 
yass: 30524 | \O�� | �<�����
                            Í�& 
```
goddamn it.
its supposed to show this:
image
^ This is a screenshot from a webGL HLMV https://danakt.com/web-hlmv/ gotta look into the code, I've already looked into it a bit, but I don't see what they're doing differently.

For example HLMV-Qt (https://github.com/MoeMod/HLMV-Qt) also only shows the "studio" 
image

I think the parser in HLMV (and derivatives) just skip it completely. 
If that ^ is so, then it would unfortunately mean that a rewrite of studio_model.cpp would probably be the best bet :| Not just for this, but for other stuff as well. Or at least a partial rewrite. And by rewrite I mean gl stuff directly in there like in HLMV and derivatives (i.e. not using the IEngineStudio).

For completeness, this is what a .qc file for the original shotgun has (taken from HLSDK (in Tools on Steam)/Weapon Models):
```
// whole body
$body studio "v_shotgun_reference"
$body studio "shell"
```

I'll see what next days bring, until then, drafting.

chinese-soup:
The origin of the shotgun is set to 3.0, therefore with 
	GLfloat _near = 3.0f;
the shell never appears, because it's "behind us" (unless you are using _ofs_forward, which obviously, well, changes the origin of the viewmodel camera backwards so...)

I've pushed that change, from the looks of it, this doesn't really break anything else other than fixing the shell, but still wanna look into it more. Maybe even get the origin of each viewmodel(?) and set that as _near for each of them, since  they all have an origin set in their .qc file...

YaLTeR:
Thanks

PR #110, opened 2021-04-22
Implement cl_old_scoreboard

### LINKED ISSUE #89: cl_old_scoreboard 1

chinese-soup:
This closes #89.

Oof, I might finally finish this. 

This is based on the original AG old scoreboard (which is based on the old HL scoreboard).
I've tried to stick as close to the original as I could but at the same time I wanted to improve a few things, one of them being rendering this somewhat accurately and be usable on Linux as well. Therefore I have made some alternative design (positioning mainly) choices. I've also introduced a `cl_oldscoreboard_width` cvar where the motivation was that since the width is already different on Linux, I might as well let the user change it to whatever he wants. Plus this also works out well for the user if the user wants to use a custom HUD font and/or a different size, they can use cl_oldscoreboard_width to make it all fit again.

So yeah, positions were changed a bit to accommodate Linux and improve upon the original, but in the end, Linux still has to have a different default width value since the HUD font is obviously bigger there and on 320 the text would overlap with the original values.

Orig:
```c++
// X positions
// relative to the side of the scoreboard
#define NAME_RANGE_MIN  20
#define NAME_RANGE_MAX  145
#define KILLS_RANGE_MIN 130
#define KILLS_RANGE_MAX 170
#define DIVIDER_POS		180
#define DEATHS_RANGE_MIN  185
#define DEATHS_RANGE_MAX  210
#define PING_RANGE_MIN	245
#define PING_RANGE_MAX	295

#define SCOREBOARD_WIDTH 320
```
This implementation:
```c++
	m_WidthScale = m_pCvarOldScoreboardWidth->value / 320.0f;

	int NAME_RANGE_MIN = 20 * m_WidthScale;
	int NAME_RANGE_MAX = 145 * m_WidthScale;
	int KILLS_RANGE_MIN = 130 * m_WidthScale;
	int KILLS_RANGE_MAX = 160 * m_WidthScale;
	int DIVIDER_POS     = 180 * m_WidthScale;
	int DEATHS_RANGE_MIN = 190 * m_WidthScale; // og 185
	int DEATHS_RANGE_MAX = 220 * m_WidthScale;
	// int PING_RANGE_MIN  = 245 * m_WidthScale;
	int PING_RANGE_MAX  = 295 * m_WidthScale;
	int END = 315 * m_WidthScale;

	int xpos_rel = (ScreenWidth - (int)m_pCvarOldScoreboardWidth->value) / 2; // scale the scoreboard based on the cvar
```

# Cvars
* **cl_oldscoreboard 0/1** (defaults to 0, therefore draws the VGUI scoreboard)
* **cl_oldscoreboard_width 320** (defaults to 320 on Windows, basically the original AG 6.6 "value"; defaults to 360 (TODO: OR 380?) on Linux)
* **cl_old_scoreboard_colortags 0/1** (defaults to 0) - disable / enable colortags rendering in nicknames

# Caveats:
* The maximum score/number of deaths that can be rendered without overlapping something else is a 4-digit number (9999) (this can be lower if you set your width really low (on Linux for example) and there's a lot of numbers that are wide (e.g. "0")). I don't see this as a problem since I have yet to see a combined team score higher than 1000 in an actual game (let alone a player score >= 1000). 
* ~~No support for color tags as I couldn't be bothered personally (because of the nickname cutoff code & I already spent way too much time than I anticipated on this) and they weren't in the original either.~~
* Cannot use localization titles.txt for **"#SCORE"** and **"#LATENCY"** since those are drawn using `DrawHudStringRightAligned` which that introduces a bug with localized strings where the X position is off by several pixels randomly (on Linux only, IIRC?) See screenshot below.
* Introduces `gHUD.DrawHudNumberStringFixed` to save some lines of code in `hud_scoreboard.cpp`. I didn't opt for replacing/removing the original DrawHudNumberString function since it could be useful in the future (basically backwards compability, even though it's broken and not used in anything right now -- the only other code which used it is already replaced by #105). I can replace it if you want me to, though.
* No idea what the positioning / sizes look like on macOS. I do have a Hackintosh on an SSD laying around here, but I think that's Catalina without 32-bit so I didn't bother, sorry. Just use the VGUI scoreboard there, I guess.
* More stuff can be read in the code comments

# Miscellaneous / features / info:
* This works rather well out-of-the-box with a custom HUD font (I tried with a monospace - Droid Sans Mono) if the user doesn't change the "tall" (font size) value. If for example the user changes the "tall" value in TrackerScheme.res to a higher value, he can  then use cl_oldscoreboard_width can fix it up so it looks good again. (See screenshots for example)
* Although positions were shifted around as mentioned before, the **basic "layout" is the same** (this includes the part where the scoreboard was never actually centered to the center of the screen), **meaning the X values for the start and end are basically the same just 'scaled' by the cl_oldscoreboard_width, where 320 is the base value.**
* Team line colors in team based modes are the same like in the original - using the team color.
* Player line colors in team based modes are the same like in the original - using the team color.
* Player line colors in non-team based modes are the same like in the original - using the hud_color.
* All the other used colors are conforming to hud_color like in the original.
* More stuff can be read in the code comments

# Screenshots

### width = 320
original_ag_66

## This implementation
**Ignore the yellow color in lines** even though it's taken in TDM mode, that's just something wrong in my HLDS kreedz gamemode setup that I have for testing (you can see the proper colors in a screenshot below all these that are mostly just to show the positioning)... 

### width = 320
**Windows:**
win_320

**Linux:** (To show why the default isn't 320 on Linux)
linux_320_why_not

### width = 380
**Windows:**
win_380

**Linux:**
linux_380

### width = 400
**Windows:**
win_400

**Linux:**
linux_400

### width = 480
**Windows:**
win_480

**Linux:**
linux_480

### width = 560
**Windows:**
win_560

**Linux:**
linux_560

### width = 640
**Windows:**
win_640

**Linux:**
linux_640

## This implementation (actual colors, no broken HLDS TDM like in the ones above)
**Linux:**
linux_380_colors

## This implementation  (Custom HUD font)
(Ignore the yellow color in lines even though it's taken in TDM mode and hud_color is set to red, that's just something wrong in my HLDS kreedz setup I have for testing...)
### Droid Sans Mono (Linux) tall = 20 (original) && cl_oldscoreboard_width = 480 @ 1280x720
droid_sans_mono_linux_tall_20_480_1080p

### Droid Sans Mono (Linux) tall = 40 (2x times original) && cl_oldscoreboard_width = 1024 @ 1920x1080
droid_sans_mono_linux_1024_1080p

--------------

**The localization bug (mentioned above) screenshot**
bufferedlocalisetextstring_bug

### DIFF
--- a/cl_dll/hud.h
+++ b/cl_dll/hud.h
@@ -99,6 +99,7 @@ struct HUDLIST {
 #include "hud_debug.h"
 #include "hud_location.h"
 #include "hud_nextmap.h"
+#include "hud_oldscoreboard.h"
 #include "hud_playerid.h"
 #include "hud_scores.h"
 #include "hud_settings.h"
@@ -605,6 +606,7 @@ public:
 	int DrawHudString(int x, int y, int iMaxX, const char *szString, int r, int g, int b );
 	int DrawHudStringReverse( int xpos, int ypos, int iMinX, const char *szString, int r, int g, int b );
 	int DrawHudNumberString( int xpos, int ypos, int iMinX, int iNumber, int r, int g, int b );
+	int DrawHudNumberStringFixed( int xpos, int ypos, int iNumber, int r, int g, int b );
 	int GetNumWidth(int iNumber, int iFlags);
 
 	int DrawHudStringCentered(int x, int y, const char* string, int r, int g, int b);
@@ -685,6 +687,7 @@ public:
 	CHudTimer		m_Timer;
 	CHudVote		m_Vote;
 	CHudWatermark	m_Watermark;
+	CHudOldScoreboard m_OldScoreBoard;
 
 	CRainbow m_Rainbow;
 
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -6,6 +6,7 @@
 #include "vgui_TeamFortressViewport.h"
 #include "hud_oldscoreboard.h"
 #include "vgui_ScorePanel.h"
+#include "steam_id.h"
 
 // Y positions
 // Those who play on killed-off 32bit Apple don't deserve old_scoreboard looking good
@@ -240,7 +241,14 @@ int CHudOldScoreboard::Draw(float fTime)
 
 			char szName[128];
 			int specoffset = 0;
-			snprintf(szName, ARRAYSIZE(szName), "%s", pl_info->name);
+			const char* name = nullptr;
+
+			if (steam_id::is_showing_real_names())
+				name = steam_id::get_real_name(scoreboard->m_iSortedRows[iRow] - 1).c_str();
+			if (!name || name[0] == '\0')
+				name = pl_info->name;
+
+			snprintf(szName, ARRAYSIZE(szName), "%s", name);
 
 			// If this player is a spectator we need to also fit " (S)" in, let's prepare for that
 			if (g_IsSpectator[scoreboard->m_iSortedRows[iRow]])
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -260,8 +260,8 @@ int CHudOldScoreboard::Draw(float fTime)
 				std::string spec_str(" (S)");
 				for( char& c : spec_str )
 				{
-					auto width = gHUD.m_scrinfo.charWidths[ static_cast<unsigned char>(c) ];
-					specoffset += width;
+					auto cwidth = gHUD.m_scrinfo.charWidths[ static_cast<unsigned char>(c) ];
+					specoffset += cwidth;
 				}
 			}
 
@@ -273,13 +273,7 @@ int CHudOldScoreboard::Draw(float fTime)
 
 			// Now that we have space for it, add the (S)
 			if (g_IsSpectator[scoreboard->m_iSortedRows[iRow]])
-			{
-				auto len = strlen(szName);
-				szName[len - 1] = ')';
-				szName[len - 2] = 'S';
-				szName[len - 3] = '(';
-				szName[len - 4] = ' ';
-			}
+				strcat(szName, " (S)");
 
 			gHUD.DrawHudStringWithColorTags(xpos + nameoffset, ypos, szName, r, g, b);
 
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -85,17 +85,12 @@ void CHudOldScoreboard::ShowScoreboard(bool bShow)
 
 int CHudOldScoreboard::Draw(float fTime)
 {
-	// Let users use 320 even on Linux, if they really want to
-	if (m_pCvarOldScoreboardWidth->value < 320.0f || m_pCvarOldScoreboardWidth->value > ScreenWidth)
-	{
-		gEngfuncs.Con_Printf("Invalid cl_oldscoreboard_width value (%d) ", (int)m_pCvarOldScoreboardWidth->value);
-		gEngfuncs.Con_Printf("(minimum = %d, maximum = %d (current screen width))\n", 320, ScreenWidth);
-		gEngfuncs.Con_Printf("Resetting to default: %s\n", DEFAULT_WIDTH);
-		gEngfuncs.Cvar_SetValue("cl_old_scoreboard_width", DEFAULT_WIDTH_NUM);
-	}
-
 	if (!IsVisible())
 		return 1;
+
+	// Let users use 320 even on Linux, if they really want to
+	int width = max(min((int)m_pCvarOldScoreboardWidth->value, ScreenWidth), 320);
+	m_WidthScale = (float)width / 320.0f;
 
 	/*
 	 * X positions, relative to the side of the scoreboard
@@ -103,8 +98,6 @@ int CHudOldScoreboard::Draw(float fTime)
 	 * And yes, the original scoreboard isn't perfectly centered,
 	 * even that is preserved :)
 	 */
-	m_WidthScale = m_pCvarOldScoreboardWidth->value / 320.0f;
-
 	int NAME_RANGE_MIN = 20 * m_WidthScale;
 	int NAME_RANGE_MAX = 145 * m_WidthScale;
 	int KILLS_RANGE_MIN = 130 * m_WidthScale;
@@ -121,7 +114,7 @@ int CHudOldScoreboard::Draw(float fTime)
 	int xpos = 0;
 	int ypos = 0;
 	float list_slot = 0;
-	int xpos_rel = (ScreenWidth - (int)m_pCvarOldScoreboardWidth->value) / 2; // scale the scoreboard based on the cvar
+	int xpos_rel = (ScreenWidth - width) / 2; // scale the scoreboard based on the cvar
 
 	// print the heading line
 	ypos = ROW_TOP + ROW_RANGE_MIN + (list_slot * ROW_GAP);
@@ -140,7 +133,7 @@ int CHudOldScoreboard::Draw(float fTime)
 	gHUD.DrawHudString(DEATHS_RANGE_MIN + xpos_rel, ypos, ScreenWidth, CHudTextMessage::BufferedLocaliseTextString("#DEATHS"), r, g, b);
 
 	// can't use #LATENCY as RightAligned is not a friend with BufferedLocaliseTextString for some reason, sorry
-	xpos = m_pCvarOldScoreboardWidth->value + xpos_rel - (m_pCvarOldScoreboardWidth->value - END);
+	xpos = width + xpos_rel - (width - END);
 	gHUD.DrawHudStringRightAligned(xpos, ypos, "Ping/loss", r, g, b);
 
 	list_slot += 1.5;
@@ -298,7 +291,7 @@ int CHudOldScoreboard::Draw(float fTime)
 
 			// draw ping & packetloss
 			// Why 25: 320 (default width in the original AG) - 295 (PING_RANGE_MAX) = 25
-			xpos = m_pCvarOldScoreboardWidth->value + xpos_rel - (m_pCvarOldScoreboardWidth->value - PING_RANGE_MAX);
+			xpos = width + xpos_rel - (width - PING_RANGE_MAX);
 			static char buf[64];
 			snprintf(buf, ARRAYSIZE(buf), "%d/%d", (int)pl_info->ping, (int)pl_info->packetloss);
 			gHUD.DrawHudStringRightAligned(xpos , ypos, buf, r, g, b);
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -29,7 +29,6 @@
 
 #define ROW_GAP (gHUD.m_scrinfo.iCharHeight - 5)
 #define ROW_RANGE_MAX ( ScreenHeight - 50 )
-#define ROW_TOP 40
 
 #define TEAM_NO             0
 #define TEAM_YES            1
@@ -85,6 +84,11 @@ int CHudOldScoreboard::Draw(float fTime)
 {
 	if (!IsVisible())
 		return 1;
+
+	// This is now calculated so that the HLKreedz timer is not in the way on ANY resolution
+	// Since the Y pos of the timer is changed with different resolution
+	// It's also calculated dynamically since one can change the ScreenHeight by resizing the window during game
+	int ROW_TOP = (ScreenHeight / 320.0f) * 30.0f;
 
 	// Let users use 320 even on Linux, if they really want to
 	int width = max(min((int)m_pCvarOldScoreboardWidth->value, ScreenWidth), 320);
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -32,15 +32,16 @@
 #define ROW_RANGE_MAX ( ScreenHeight - 50 )
 #define ROW_TOP 40
 
-#define TEAM_NO				0
-#define TEAM_YES			1
-#define TEAM_SPECTATORS		2
-#define TEAM_BLANK			3
+#define TEAM_NO             0
+#define TEAM_YES            1
+#define TEAM_SPECTATORS     2
+#define TEAM_BLANK          3
 
 int CHudOldScoreboard::Init(void)
 {
 	m_pCvarOldScoreboard = CVAR_CREATE("cl_old_scoreboard", "0", FCVAR_ARCHIVE);
 	m_pCvarOldScoreboardWidth = CVAR_CREATE("cl_old_scoreboard_width", DEFAULT_WIDTH, FCVAR_ARCHIVE);
+	m_pCvarOldScoreboardColorsTags = CVAR_CREATE("cl_old_scoreboard_colortags", "0", FCVAR_ARCHIVE);
 
 	m_iFlags = 0;
 
@@ -84,7 +85,6 @@ void CHudOldScoreboard::ShowScoreboard(bool bShow)
 
 int CHudOldScoreboard::Draw(float fTime)
 {
-	gEngfuncs.Con_Printf("m_iFlags = %d\n", m_iFlags);
 	// Let users use 320 even on Linux, if they really want to
 	if (m_pCvarOldScoreboardWidth->value < 320.0f || m_pCvarOldScoreboardWidth->value > ScreenWidth)
 	{
@@ -249,10 +249,8 @@ int CHudOldScoreboard::Draw(float fTime)
 			}
 
 			char szName[128];
-			char colorlessName[128];
 			int specoffset = 0;
 			snprintf(szName, ARRAYSIZE(szName), "%s", pl_info->name);
-			color_tags::strip_color_tags(colorlessName, szName, ARRAYSIZE(colorlessName));
 
 			// If this player is a spectator we need to also fit " (S)" in, let's prepare for that
 			if (g_IsSpectator[scoreboard->m_iSortedRows[iRow]])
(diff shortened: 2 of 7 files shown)


chinese-soup:
I suppose this is now ready for da grillin'.

cl_dll/hud.h
```diff
@@ -683,9 +685,11 @@ class CHud
 	CHudTimer		m_Timer;
 	CHudVote		m_Vote;
 	CHudWatermark	m_Watermark;
+	CHudOldScoreboard m_OldScoreBoard;
 
 	CRainbow m_Rainbow;
 
+
```
YaLTeR:
```suggestion
```

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	 * since it doesn't look good (Ping/loss doesn't fit and overlaps "Deaths" etc.)
+	 * on the (original) 320 default, like it does on Windows,
+	 * since the engine's Trebuchet MS is bigger than on Windows
+	 * WINDOWS = Trebuchet MS = 9x8
+	 * LINUX = Trebuchet MS = 11x10
+	 */
+	#define DEFAULT_WIDTH "380"
+	#define DEFAULT_WIDTH_NUM 380
+#else
+	#define ROW_GAP  (gHUD.m_scrinfo.iCharHeight - 5)
```
YaLTeR, line 27: Is this needed if it's the same in both?

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+#define TEAM_YES			1
+#define TEAM_SPECTATORS		2
+#define TEAM_BLANK			3
+
+int CHudOldScoreboard::Init(void)
+{
+	m_pCvarOldScoreboard = CVAR_CREATE("cl_old_scoreboard", "0", FCVAR_ARCHIVE);
+	m_pCvarOldScoreboardWidth = CVAR_CREATE("cl_old_scoreboard_width", DEFAULT_WIDTH, FCVAR_ARCHIVE);
+
+	m_bShowScoreboard = false;
```
YaLTeR, line 46: Isn't this what `m_iFlags & HUD_ACTIVE` is for? Why duplicate that with a boolean?

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+int CHudOldScoreboard::Draw(float fTime)
+{
+	// Let users use 320 even on Linux, if they really want to
+	if (m_pCvarOldScoreboardWidth->value < 320.0f || m_pCvarOldScoreboardWidth->value > ScreenWidth)
+	{
+		gEngfuncs.Con_Printf("Invalid cl_oldscoreboard_width value (%d) ", (int)m_pCvarOldScoreboardWidth->value);
+		gEngfuncs.Con_Printf("(minimum = %d, maximum = %d (current screen width)\n", 320, ScreenWidth);
+		gEngfuncs.Con_Printf("Resetting to default: %s\n", DEFAULT_WIDTH);
+		m_pCvarOldScoreboardWidth->value = (float)DEFAULT_WIDTH_NUM;
```
YaLTeR, line 93: Does that actually update the cvar string properly?

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	m_bShowScoreboard = bShow;
+}
+
+int CHudOldScoreboard::Draw(float fTime)
+{
+	// Let users use 320 even on Linux, if they really want to
+	if (m_pCvarOldScoreboardWidth->value < 320.0f || m_pCvarOldScoreboardWidth->value > ScreenWidth)
+	{
+		gEngfuncs.Con_Printf("Invalid cl_oldscoreboard_width value (%d) ", (int)m_pCvarOldScoreboardWidth->value);
+		gEngfuncs.Con_Printf("(minimum = %d, maximum = %d (current screen width)\n", 320, ScreenWidth);
```
YaLTeR, line 91:
```suggestion
		gEngfuncs.Con_Printf("(minimum = %d, maximum = %d (current screen width))\n", 320, ScreenWidth);
```

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+{
+	if (bShow && gViewPort && gViewPort->m_pScoreBoard)
+		gViewPort->m_pScoreBoard->RebuildTeams();
+	m_bShowScoreboard = bShow;
+}
+
+int CHudOldScoreboard::Draw(float fTime)
+{
+	// Let users use 320 even on Linux, if they really want to
+	if (m_pCvarOldScoreboardWidth->value < 320.0f || m_pCvarOldScoreboardWidth->value > ScreenWidth)
```
YaLTeR, line 88: A cvar that resets if you accidentally launch game in wrong res sounds really annoying

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	m_IconFlagScore.rc = gHUD.GetSpriteRect(m_iSprite);
+	m_WidthScale = m_pCvarOldScoreboardWidth->value / 320.0f;
+	m_bShowScoreboard = false;
+
+	return 1;
+}
+
+void CHudOldScoreboard::Reset(void)
+{
+
```
YaLTeR, line 70:
```suggestion
```

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	m_WidthScale = m_pCvarOldScoreboardWidth->value / 320.0f;
+
+	int NAME_RANGE_MIN = 20 * m_WidthScale;
+	int NAME_RANGE_MAX = 145 * m_WidthScale;
+	int KILLS_RANGE_MIN = 130 * m_WidthScale;
+	int KILLS_RANGE_MAX = 160 * m_WidthScale;
+	int DIVIDER_POS     = 180 * m_WidthScale;
+	int DEATHS_RANGE_MIN = 190 * m_WidthScale; // og 185
+	int DEATHS_RANGE_MAX = 220 * m_WidthScale;
+	// int PING_RANGE_MIN  = 245 * m_WidthScale;
```
YaLTeR, line 114: No need to leave commented out code

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	ypos = ROW_TOP + ROW_RANGE_MIN + (list_slot * ROW_GAP);
+	xpos = NAME_RANGE_MIN + xpos_rel;
+
+	if (!gHUD.m_Teamplay)
+		gHUD.DrawHudString(xpos, ypos, NAME_RANGE_MAX + xpos_rel, CHudTextMessage::BufferedLocaliseTextString("#PLAYERS"), r, g, b);
+	else
+		gHUD.DrawHudString(xpos, ypos, NAME_RANGE_MAX + xpos_rel, CHudTextMessage::BufferedLocaliseTextString("#TEAMS"), r, g, b);
+
+	// can't use #SCORE as RightAligned is not a friend with BufferedLocaliseTextString for some reason, sorry
+	gHUD.DrawHudStringRightAligned( KILLS_RANGE_MAX + xpos_rel, ypos, CHudTextMessage::BufferedLocaliseTextString("Score"), r, g, b);
```
YaLTeR, line 136:
```suggestion
	gHUD.DrawHudStringRightAligned(KILLS_RANGE_MAX + xpos_rel, ypos, CHudTextMessage::BufferedLocaliseTextString("Score"), r, g, b);
```

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+				r = g = b = 255;
+				// overlay the background in blue, then draw the score text over it
+				FillRGBA( NAME_RANGE_MIN + xpos_rel - 5, ypos + 5, PING_RANGE_MAX - 5, ROW_GAP, 0, 0, 255, 70);
+			}
+
+			char szName[128];
+			char colorlessName[128];
+			int specoffset = 0;
+			snprintf(szName, ARRAYSIZE(szName), "%s", pl_info->name);
+			color_tags::strip_color_tags(colorlessName, szName, ARRAYSIZE(colorlessName));
```
YaLTeR, line 255: Maybe we want to draw names in color?

cl_dll/hud_oldscoreboard.h
```diff
@@ -0,0 +1,36 @@
...
+
+	icon_flagstatus_t m_IconFlagScore;
+
+	bool m_bShowScoreboard;
+
+	cvar_t *m_pCvarOldScoreboard;
+	cvar_t *m_pCvarOldScoreboardWidth;
+
+	float m_WidthScale;
+
```
YaLTeR, line 33:
```suggestion
```

YaLTeR (changes requested):

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Changed it around couple times and then forgot it there, thanks.

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Right, it only updated the "internal" value, not the value shown to user. Fixed with `gEngfuncs.Cvar_SetValue("cl_old_scoreboard_width", DEFAULT_WIDTH_NUM);`

This obviously still only changes the value after the user is back in the game as HUD Draw() functions aren't called in the menu (where you use the console).

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Do you reach an actual game when in that "accidental resolution" though? What do you suggest, just cap it above 320 and get rid of the high cap?

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+	gHUD.DrawHudString(DEATHS_RANGE_MIN + xpos_rel, ypos, ScreenWidth, CHudTextMessage::BufferedLocaliseTextString("#DEATHS"), r, g, b);
+
+	// can't use #LATENCY as RightAligned is not a friend with BufferedLocaliseTextString for some reason, sorry
+	xpos = m_pCvarOldScoreboardWidth->value + xpos_rel - (m_pCvarOldScoreboardWidth->value - END);
+	gHUD.DrawHudStringRightAligned(xpos, ypos, "Ping/loss", r, g, b);
+
+	list_slot += 1.5;
+	ypos = ROW_TOP + ROW_RANGE_MIN + (list_slot * ROW_GAP);
+	xpos = NAME_RANGE_MIN + xpos_rel;
+	FillRGBA( xpos, ypos, PING_RANGE_MAX, 1, r, g, b, 255); // draw the seperator line
```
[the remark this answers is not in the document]
chinese-soup: Legacy code being copied over and then not fixed up. :)

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+			// draw deaths
+			xpos = DEATHS_RANGE_MAX + xpos_rel;
+			gHUD.DrawHudNumberStringFixed(xpos, ypos, pl_info_extra->deaths, r, g, b);
+
+			// draw ping & packetloss
+			// Why 25: 320 (default width in the original AG) - 295 (PING_RANGE_MAX) = 25
+			xpos = m_pCvarOldScoreboardWidth->value + xpos_rel - (m_pCvarOldScoreboardWidth->value - PING_RANGE_MAX);
+			static char buf[64];
+			sprintf(buf, "%d/%d", (int)pl_info->ping, (int)pl_info->packetloss);
```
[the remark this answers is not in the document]
chinese-soup: Sry, legacy copied over again.

cl_dll/vgui_TeamFortressViewport.cpp
```diff
@@ -1429,6 +1447,14 @@ void TeamFortressViewport::HideScoreBoard( void )
 	if ( gHUD.m_iIntermission )
 		return;
 
+	// The user can change the scoreboard style while the scoreboard is being "drawn" (e.g. when togglescores is on)
+	// Therefore hide both of them so that we are sure we aren't drawing both of them
+
+	if ( CVAR_GET_FLOAT("cl_old_scoreboard") == 1 )
```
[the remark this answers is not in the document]
chinese-soup: Um, the function starts with
```c++
void TeamFortressViewport::HideScoreBoard( void )
{
	// Prevent removal of scoreboard during intermission
	if ( gHUD.m_iIntermission )
		return;
```
Why should it hide during an intermission?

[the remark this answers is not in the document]
chinese-soup: Oh right, I am smart. I guess you mean after the intermission is over... Well, it "doesn't hide", but it won't show up either, because after an intermission the map changes and VidInit() is called again?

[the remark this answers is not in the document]
chinese-soup: Anyhow, removed the check altogether.

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,309 @@
...
+				colorlessName[strlen(colorlessName) - 1] = '\0';
+			}
+
+			// Now that we have space for it, add the (S)
+			if (g_IsSpectator[scoreboard->m_iSortedRows[iRow]])
+			{
+				colorlessName[strlen(colorlessName) - 1] = ')';
+				colorlessName[strlen(colorlessName) - 2] = 'S';
+				colorlessName[strlen(colorlessName) - 3] = '(';
+				colorlessName[strlen(colorlessName) - 4] = ' ';
```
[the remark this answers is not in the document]
chinese-soup: Right. :smiley: Is this better?

cl_dll/hud_oldscoreboard.cpp
chinese-soup: From the PR text:
> No support for color tags as I couldn't be bothered personally (because of the nickname cutoff code & I already spent way too much time than I anticipated on this) and they weren't in the original either.

The names use hud_color and in TDM modes team color.

I had it in mind as a toggleable cvar (I know, I know), since I from what I've gathered this feature is for the hardcore AG 6.6 (Alex comes in mind) players who want the legacy stuff and the old scoreboard never did support colortags (nor does the VGUI scorepanel). 

I then removed it altogether since I couldn't be bothered to make it work for both, since it requires not-as-trivial-as-is-right-now solution to cutting off the nicknames.

I could (except for the `(S)` part) make it work with just colortags always enabled, but then in TDM modes the "white" (non-colortagged) letters will be the team font which breaks the "color tag" principle. (Since it should be the hud_color font) - e.g. really^2LongNickname

Anyway, if you really think colortags are a priority, I'll go ahead and implement them, but personally I'd still rather cvar it.

Here's an older screenshot (from like 25th of April, obviously the positions are all off etc.) with color tags for illustration of awfulness (IMO, anyway):

crossfire0065

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Legacy by Valve again, sorry. Fixed.

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Disregard (almost) everything I said, I am retarded & didn't realize there's already a `GetHudStringWidthWithColorTags` function which makes this easier. I still insist on this being a cvar based on the arguments I provided, pushing a version with a cvar and immediately asking:

Is this OK or can this blow up?
```
color_tags::strip_color_tags(szName, szName, ARRAYSIZE(szName));
```

chinese-soup:
I've added a cvar for toggling colortags in nicknames. I'd prefer it to be a cvar based on the argument I've posted in the review.

Also fixed `specoffset` not being used at all accidentally.

chinese-soup:
Fixed accidental doubled call of `UpdateCursorState();`

cl_dll/hud_oldscoreboard.cpp
YaLTeR: Hm, could you ask ppl who want the feature if they want color tags or no color tags? I think it makes the most sense to have only the behavior they want, no need for a toggle.

cl_dll/hud_oldscoreboard.cpp
YaLTeR: I think at least doing a min(value, screen width) and max(value, 320) in the code without changing the actual cvar value is better.

cl_dll/hud_oldscoreboard.h
```diff
@@ -0,0 +1,34 @@
...
+class CHudOldScoreboard: public CHudBase
+{
+public:
+	int Init( void );
+	int VidInit( void );
+	int Draw( float flTime );
+	void Reset( void );
+
+	bool IsVisible( );
+	void ShowScoreboard( bool bShow = true );
```
YaLTeR, line 17:
```suggestion
	void ShowScoreboard( bool bShow );
```

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Never mind, colortags in oldscoreboard are default in AG 6.6, I just misremembered, sorry. Smiley and Alex said to leave them in by default like in AG 6.6, so I've removed the cvar as per this discussion. :)

chinese-soup:
Ready for re-review.
* removed colortags cvar (colortags are always on)
* min-max'd the width cvar 
* removed the default value for `bool bShow = true`

cl_dll/hud_oldscoreboard.cpp
```diff
@@ -0,0 +1,300 @@
...
+	 * since the engine's Trebuchet MS is bigger than on Windows
+	 * WINDOWS = Trebuchet MS = 9x8
+	 * LINUX = Trebuchet MS = 11x10
+	 */
+	#define DEFAULT_WIDTH "380"
+	#define DEFAULT_WIDTH_NUM 380
+#else
+	#define ROW_RANGE_MIN (gHUD.m_scrinfo.iCharHeight + 2)
+	#define DEFAULT_WIDTH "320"
+	#define DEFAULT_WIDTH_NUM 320
```
YaLTeR, line 28: As far as I can see the _NUM is no longer used.

YaLTeR (commented):
Alright, now just need to get someone else to give this a test.

YaLTeR (approved):

cl_dll/hud_oldscoreboard.cpp
chinese-soup: Correct. Got rid of it.

execut4ble:
> 
> 
> Alright, now just need to get someone else to give this a test.

I tested it today, seems to work fine, no realnames support yet but I think shoup intends to add that later?
I also thought it would be cool to have a CVAR to vertically offset the scoreboard from the top of the screen, but that can be done as a later improvement.

Other than that it works and looks good to me.

chinese-soup:
> no realnames support yet but I think shoup intends to add that later?
I've added the realnames support, I just forgot that it was a thing.

> I also thought it would be cool to have a CVAR to vertically offset the scoreboard from the top of the screen, but that can be done as a later improvement.
Depends on what @YaLTeR thinks of this. It'd help for stuff like the HLKreedz timer being in the way (there's nothing in OpenAG that's in the way, only HUD elements that could can be in the way are the 3rd party servers ones...).

kz_satomi0001

YaLTeR:
> Depends on what @YaLTeR thinks of this. It'd help for stuff like the HLKreedz timer being in the way of the scoreboard header (there's nothing in OpenAG that's in the way, only HUD elements that could can be in the way are the 3rd party servers ones...).

Maybe move it further down so it's not obstructed by the timer?

chinese-soup:
> Maybe move it further down so it's not obstructed by the timer?

This only solves it for the timer and goes against the "as close to the AG 6.6 implementation as we can get", but alright.

Done. Tested by me on Linux, with several random resolutions changing the window size and exec on Windows @ 1080p.

image

720p


```c++
	// This is now calculated so that the HLKreedz timer is not in the way on ANY res
	// Since both the Y pos of both the timer and the scoreboard are changed with different resolution
	// It's also calculated dynamically since one can change the ScreenHeight by resizing the window during game
	int ROW_TOP = (ScreenHeight / 320.0f) * 30.0f;
```

chinese-soup:
* Fixed a bug where nicknames of spectating players were cut off twice for the " (S)" addition.

YaLTeR (approved):

YaLTeR:
Thanks!

PR #111, opened 2021-05-13
Fix viewmodel attachment positions

tmp64:
Fixes incorrect weapon effect positions (sprites and beams) when viewmodel FOV is different from default. It uses code from Source SDK (c_baseviewmodel.cpp, ::FormatViewModelAttachment) to reproject on-screen position from main view FOV to viewmodel FOV.

I used global variables to save view position and direction vectors since they are not available outside of `V_CalcRefDef`.

Demo:
https://www.youtube.com/watch?v=9_10Oms73Fg


YaLTeR:
Thanks!

PR #112, opened 2021-05-19
Add an option to enable printing chat messages to console immediately

chinese-soup:
Hello.

This is a thing that bugged me for a while, since I sometimes I really want to copy a chat message immediately (to google translate for example :^), but at the same time I prefer using a hud_saytext_time above 5 seconds for normal chatting. 

This way it won't wait for the ScrollTextUp if `hud_saytext_immediate_print` is set to 1. I also added the archive flag to the **hud_saytext_time** cvar.

Also, is the cvar name alright? Seems a bit long to me, but I can't think of anything descriptive and at the same time shorter. (instant_print? print_instantly? ...)

### DIFF
--- a/cl_dll/saytext.cpp
+++ b/cl_dll/saytext.cpp
@@ -19,6 +19,7 @@
 //
 
 #include "hud.h"
+#include <ctime>
 #include "cl_util.h"
 #include "parsemsg.h"
 
@@ -57,7 +58,7 @@ int CHudSayText :: Init( void )
 	InitHUDData();
 
 	m_HUD_saytext =			gEngfuncs.pfnRegisterVariable( "hud_saytext", "1", 0 );
-	m_HUD_saytext_time =	gEngfuncs.pfnRegisterVariable( "hud_saytext_time", "5", 0 );
+	m_HUD_saytext_time =	gEngfuncs.pfnRegisterVariable( "hud_saytext_time", "5", FCVAR_ARCHIVE );
 	m_HUD_saytext_sound =	gEngfuncs.pfnRegisterVariable( "hud_saytext_sound", "0", FCVAR_ARCHIVE );
 
 	m_iFlags |= HUD_INTERMISSION; // is always drawn during an intermission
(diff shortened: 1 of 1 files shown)


YaLTeR:
Does this need to be an option? I honestly never understood why it printed to console only when the message is hidden, it always seemed more annoying than anything.

YaLTeR:
Yeah

chinese-soup:
done

tmp64:
Can you also add a timestamp to the beginning so the message can be found more easily in the console?
```
random console junk
Player killed ^1A^5G^6-^4Pro with pure skel
[15:32:59] Chat message text
Player disconnected (kicked)
```

chinese-soup:
> Can you also add a timestamp to the beginning so the message can be found more easily in the console?
> 
> ```
> random console junk
> Player killed ^1A^5G^6-^4Pro with pure skel
> [15:32:59] Chat message text
> Player disconnected (kicked)
> ```

Sure, I was thinking of that yesterday and then I completely forgot. Should this be an option or should it be always on (or should there be an option for strftime()-like formatting?)

YaLTeR:
no options pls

tmp64:
I don't think there is a need to make it an option and it should be always on.

chinese-soup:
How's this?

cl_dll/saytext.cpp
```diff
@@ -221,6 +221,20 @@ void CHudSayText :: SayTextPrint( const char *pszBuf, int iBufSize, int clientIn
 		ConsolePrint( pszBuf );
 		return;
 	}
+	char timestamp[16];
+	std::time_t curtime = std::time(nullptr);
+	auto written = std::strftime(timestamp, sizeof(timestamp), "[%H:%M:%S] ", std::localtime(&curtime));
+
+	if(written > 0)
```
tmp64, line 228: You can just call ConsolePrint twice instead of concatenating the string
```cpp
ConsolePrint( timestamp );
ConsolePrint( pszBuf );
```

chinese-soup: oh yea, thanks

chinese-soup:
image

cl_dll/saytext.cpp
```diff
@@ -221,6 +221,14 @@ void CHudSayText :: SayTextPrint( const char *pszBuf, int iBufSize, int clientIn
...
+	    ConsolePrint( timestamp );
+
+	ConsolePrint( pszBuf );
```
YaLTeR, line 231: I see there's a `convert_locations` call below that should probably happen before the console print, maybe something else too?

chinese-soup: image

Should be fine now.

YaLTeR (approved):

YaLTeR:
Thanks!

PR #113, opened 2021-05-21
Implement command customtimer

### LINKED ISSUE #64: Customtimer
I would like to ask to add AG6.6 command called "customtimer" to openAG.

chinese-soup:
Closes #64. Implemented mostly 1:1 to the the AG 6.6 implementation. I added a bar (and a colon in the text), because I think it looks OK, but this is of course up for discussion, for example if @chortex who originally reported this issue is still around, or any other players that use customtimer in their binds.

I've also decided for showing "Timer: 0" when the time is <1.0 (and 9 when "customtimer 10" starts and number instantly becomes 9.999999) like the original implementation does instead of subtracting and adding 1.

**AG 6.6:**
image

**OpenAG:**
**GIF:** 
lal
**Video:** 
https://user-images.githubusercontent.com/5108747/119066552-4e282a00-b9e0-11eb-9db7-8e586347dac9.mp4

(The sound like in the original is present I just didn't record it.)

I've also added a check to fail if the time in seconds is greater than 2 days since in the original implementation you could overflow. I've opted for 2 days as that's the maximum `mp_timelimit` (IIRC) so there shouldn't be a need for a higher value.

### DIFF
--- a/cl_dll/hud_customtimer.cpp
+++ b/cl_dll/hud_customtimer.cpp
@@ -1,0 +1,61 @@
+#include "hud.h"
+#include "cl_util.h"
+
+DECLARE_COMMAND(m_CustomTimer, CustomTimer);
+
+int CHudCustomTimer::Init()
+{
+	HOOK_COMMAND("customtimer", CustomTimer);
+
+	m_iFlags = 0;
+
+	gHUD.AddHudElem(this);
+	return 0;
+}
+
+int CHudCustomTimer::VidInit()
+{
+	m_iFlags &= ~HUD_ACTIVE;
+	return 1;
+}
+
+int CHudCustomTimer::Draw(float time)
+{
+	if (gHUD.m_flTime >= m_flTurnoffTime || gHUD.m_iIntermission)
+	{
+		m_iFlags &= ~HUD_ACTIVE;
+		gEngfuncs.pfnPlaySoundByName("fvox/bell.wav", 1.0f);
+		return 0;
+	}
+
+	char str[64];
+	int r, g, b;
+	auto diff = m_flTurnoffTime - gHUD.m_flTime;
+	UnpackRGB(r, g, b, gHUD.m_iDefaultHUDColor);
+
+	FillRGBA((ScreenWidth / 2) - 50, (gHUD.m_scrinfo.iCharHeight * 4) - 6, (diff / m_flSeconds) * 100, 4, r, g, b, 210);
+	snprintf(str, sizeof(str), "Timer: %d", (int)diff);
+	gHUD.DrawHudStringCentered(ScreenWidth / 2, gHUD.m_scrinfo.iCharHeight * 4, str, r, g, b);
+
+	return 0;
+}
+
+int CHudCustomTimer::UserCmd_CustomTimer()
+{
+	if (gEngfuncs.Cmd_Argc() == 2)
+	{
+		auto time = atof(gEngfuncs.Cmd_Argv(1));
+		if (time > 172800)
+		{
+			gEngfuncs.Con_Printf("Time can't be higher than 172800 seconds (2 days).\n");
+			return 1;
+		}
+		m_flSeconds = time;
+		m_flTurnoffTime = gHUD.m_flTime + time;
+		m_iFlags |= HUD_ACTIVE;
+	}
+	else
+		gEngfuncs.Con_Printf("customtimer <seconds> - start a local countdown timer\n");
+
+	return 1;
+}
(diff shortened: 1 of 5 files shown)


cl_dll/hud_customtimer.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+	m_iFlags &= ~HUD_ACTIVE;
+	m_iTimerPos = gHUD.m_scrinfo.iCharHeight * 4;
+
+	return 1;
+}
+
+int CHudCustomTimer::Draw(float time)
+{
+	if (gHUD.m_flTime >= m_flTurnoffTime  || gHUD.m_iIntermission) {
```
YaLTeR, line 26:
```suggestion
	if (gHUD.m_flTime >= m_flTurnoffTime || gHUD.m_iIntermission) {
```

cl_dll/hud_customtimer.cpp
```diff
@@ -0,0 +1,61 @@
...
+		gEngfuncs.pfnPlaySoundByName("fvox/bell.wav", 1);
+		return 0;
+	}
+
+	char str[64];
+	int r, g, b;
+	UnpackRGB(r, g, b, gHUD.m_iDefaultHUDColor);
+
+	FillRGBA((ScreenWidth / 2) - 50, m_iTimerPos - 8, (-(gHUD.m_flTime - m_flTurnoffTime) / m_flSeconds) * 100, 4, r, g, b, 210);
+	sprintf(str, "Timer: %d", (int)(m_flTurnoffTime - gHUD.m_flTime));
```
YaLTeR, line 37: snprintf just in case

cl_dll/hud_customtimer.cpp
```diff
@@ -0,0 +1,61 @@
...
+		m_iFlags &= ~HUD_ACTIVE;
+		gEngfuncs.pfnPlaySoundByName("fvox/bell.wav", 1);
+		return 0;
+	}
+
+	char str[64];
+	int r, g, b;
+	UnpackRGB(r, g, b, gHUD.m_iDefaultHUDColor);
+
+	FillRGBA((ScreenWidth / 2) - 50, m_iTimerPos - 8, (-(gHUD.m_flTime - m_flTurnoffTime) / m_flSeconds) * 100, 4, r, g, b, 210);
```
YaLTeR, line 36: This doesn't overlap any other HUD elements, right?

YaLTeR, line 36:
```suggestion
	FillRGBA((ScreenWidth / 2) - 50, m_iTimerPos - 8, ((m_flTurnoffTime - gHUD.m_flTime) / m_flSeconds) * 100, 4, r, g, b, 210);
```

cl_dll/hud_customtimer.h
```diff
@@ -0,0 +1,23 @@
+#ifndef CUSTOMTIMER_H
+#define CUSTOMTIMER_H
+
+#pragma once
+#include <cstdint>
+
+
```
YaLTeR, line 7:
```suggestion
```

cl_dll/hud_customtimer.h
```diff
@@ -0,0 +1,23 @@
...
+	float m_iTimerPos;
+
+public:
+	virtual int Init();
+	virtual int VidInit();
+	virtual int Draw(float time);
+
+	int UserCmd_CustomTimer();
+};
+
```
YaLTeR, line 21:
```suggestion
```

cl_dll/hud_customtimer.cpp
```diff
@@ -0,0 +1,61 @@
...
+	m_iFlags = 0;
+
+	gHUD.AddHudElem(this);
+	return 0;
+}
+
+int CHudCustomTimer::VidInit()
+{
+	m_iFlags &= ~HUD_ACTIVE;
+	m_iTimerPos = gHUD.m_scrinfo.iCharHeight * 4;
```
YaLTeR, line 19: any reason why this is saved into a member variable? and not just computed inside Draw()

YaLTeR (changes requested):

cl_dll/hud_customtimer.cpp
chinese-soup: It doesn't at the moment, since there's no HUD element in the center that's using `iCharHeight * 3` in Open AG. In the original AG 6.6 there's a map name HUD element that appears after joining a server that'd overlap, but we don't have that in OpenAG.

```
AgDrawHudStringCentered(ScreenWidth / 2 , gHUD.m_scrinfo.iCharHeight*3, ScreenWidth,AgMapname().c_str(),r,g,b);
```

chinese-soup:
Should be all fixed up now. I also changed the position of the bar to -6 instead of -8 to make it feel less "distant".

YaLTeR (approved):

YaLTeR:
Thanks!

chortex:
Thank you guys. Absolutely like it, Idea with bar is amazing.

PR #117, opened 2021-06-02
Use the target player's weapon crosshair when spectating in first person, don't replace it

chinese-soup:
As per Alex's bug report:
> There is just 1 more thing that makes ag6.6 different: you can see your default handgun crosschair @ spectate mode in OpenAG. A lot of peeps use the command r_drawviewmodel 0 so there is no way to understand what the weapon in hand atm.


YaLTeR:
Thanks!

PR #118, opened 2021-06-03
Rainbow HUD: Fix invisible text being visible

tmp64:
```cpp
int CHud::GetHudStringWidth(const char* string)
{
	return gEngfuncs.pfnDrawString(0, 0, string, 0, 0, 0);
}
```

`GetHudStringWidth` gets the string length by drawing it with a black color (due to additive blending mode, it becomes transparent). Rainbow HUD overrides this color, causing the text to become visible.

This PR also fixes a small issue in `DrawStringReverse`. It used to call `DrawString` to get the string width. `DrawString` needs to decode the input string into separate UTF-8 chars to draw them with different colors but it isn't needed in this case. I replaced it with a call to `CHud::GetHudStringWidth`.


YaLTeR:
Thanks

PR #120, opened 2021-06-04
Skip NULL nicknames in the sorted scoreboard in order to get rid of ghost players when they disconnect while +showscores is held

### LINKED ISSUE #119: Old Scoreboard doesn't update when a player leaves
Old Scoreboard doesn't update after a player leaves while you keep it open via the +showscores command causing them to show as (null) in the scoreboard

chinese-soup:
Fixes #119 (that's present on some servers), hopefully. ;-)

**EDIT: Tested by me & @iCHOPPERi on Ironanjuu & `AGHLDM DE #2 | 1000 FPS | FastDL`, seems like this does the trick.** :)

### DIFF
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -217,6 +217,9 @@ int CHudOldScoreboard::Draw(float fTime)
 			hud_player_info_t* pl_info = &g_PlayerInfoList[sorted];
 			extra_player_info_t* pl_info_extra = &g_PlayerExtraInfo[sorted];
 
+			if (pl_info->name == NULL)
+				continue;
+
 			ypos = ROW_TOP + ROW_RANGE_MIN + (list_slot * ROW_GAP);
 
 			// check we haven't drawn too far down
--- a/cl_dll/hud_oldscoreboard.cpp
+++ b/cl_dll/hud_oldscoreboard.cpp
@@ -218,7 +218,10 @@ int CHudOldScoreboard::Draw(float fTime)
 			extra_player_info_t* pl_info_extra = &g_PlayerExtraInfo[sorted];
 
 			if (pl_info->name == NULL)
+			{
+				gViewPort->m_pScoreBoard->RebuildTeams();
 				continue;
+			}
 
 			ypos = ROW_TOP + ROW_RANGE_MIN + (list_slot * ROW_GAP);
 


YaLTeR:
Is it not taxing to do that? What does the regular scoreboard do?

chinese-soup:
It shouldn't be taxing really, the regular scoreboard just calls it whenever Msgfunc_TeamInfo()/Msgfunc_Spectator/Msgfunc_Teamscore() comes in. And also when the scoreboard is open (like we do). So literally the same way.

However, I've realized the way it probably works in the VGUI one is it just checks if the name is NULL and skips that line (That's what RebuildTeams does in like the at beginning, lol!), so let's just do that. :) It's gonna "rotate" the slot away after RebuildTeams is called anyway, so in the meantime just skip the ghost player in the list. 

```c++
			if (pl_info->name == NULL)
				continue;
```

chinese-soup:
Anyway, the vgui_ScorePanel does the same thing basically, skips NULL nicknames and therefore since it doesn't `RebuildTeams()` while open, sometimes (on some servers?) _this_ happens on both old scoreboard (after this NULL skip fix) and the current VGUI scoreboard:

vengeance0001

Because the "players" count of a team (`g_TeamInfo`) is only updated once RebuildTeams() resets all teams to 0 and then goes through all the players and checks if & what team they are on.

So I dunno if that's a thing you would want to fix as well, since again - this would require to do basically the entire RebuildTeams() call except for the "Update();" part (which updates the VGUI panel, not needed to call in old_scoreboard).

My personal opinion is to just not do the RebuildTeams() on every Draw and leave the empty teams (like in screenshot) there until the next RebuildTeams happens (which is when someone joins/changes team/changes to spec/when you reopen the scoreboard). 

What do you think?

YaLTeR:
Maybe it's possible to somehow call `RebuildTeams()` when a player leaves?

chinese-soup:
That is already happening, but isn't enough for some reason on some servers (and for the empty team not disappearing it looks like on any server), like I've already said in the original post. I assume maybe because of a race condition or something.

Server **only** sends this to the client everytime **anyone** leaves (even actual non-spectating players)
```c++
      // Tell all clients this player isn't a spectator anymore
      MESSAGE_BEGIN( MSG_ALL, gmsgSpectator );  
        WRITE_BYTE( ENTINDEX(pPlayer->edict()) );
        WRITE_BYTE( 0 );
      MESSAGE_END();
```

Which is handled by:
```c++
int TeamFortressViewport::MsgFunc_Spectator( const char *pszName, int iSize, void *pbuf )
{
	BEGIN_READ( pbuf, iSize );

	short cl = READ_BYTE();
	if ( cl > 0 && cl <= MAX_PLAYERS )
	{
		g_IsSpectator[cl] = READ_BYTE();
	}

	gEngfuncs.Con_Printf("MsgFunc_Spectator()\n");
	m_pScoreBoard->RebuildTeams();

	return 1;
}
```

As you can see in the log below, even after `RebuildTeams` is called and OldScoreboard::Draw() is accessing the scoreboard (`m_pScoreboard`) and `GetAllPlayersInfo()` is being called, on every draw there's still a team left:
(red has 1 player - me, blue has 1 bot):

```
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
TEST blue | count = 1
[11:20:42] ADMIN ^_^: kick HLKZ Dummy 4826 // Kick the last player in the blue team
[AMXX] Client "HLKZ Dummy 4826" kicked
HLKZ Dummy 4826 left with score 0
[11:20:42] - HLKZ Dummy 4826 has left the game
MsgFunc_Spectator() <--- MsgFunc_Spectator we got from the server is called
rebuildteams() <--- RebuildTeams is called at the end of MsgFunc_Spectator
TEST red | count = 1
TEST blue | count = 1  // <---  blue team is still there
TEST red | count = 1
TEST blue | count = 1
TEST red | count = 1
rebuildteams() // <--- I closed and reopened the scoreboard here.
TEST red | count = 1
TEST red | count = 1
TEST red | count = 1
```

chinese-soup:
Like I've mentioned in our Discord conversation, since there doesn't have to be a NULL player nickname & since gEngfuncs.pfnGetPlayerInfo() is slow and the player isn't dead yet when the rebuild happens the "empty team not disappearing" bug is hard to fix, only thing I can think of is starting a "timer" (save currentGameTime and wait for currentGameTime+3) after MsgFunc_Spectator is called and in 2-3 seconds call RebuildTeams again. 

I don't really see any other way to fix this. The next-generation server DLL @fireblizzard is working at should probably add a message to the disconnect `void AgGameRules::ClientDisconnected( edict_t *pClient )` that actually tells us that the player disconnected, not just that the player is no longer a spectator, like I've mentioned on Discord.

```c++
void AgGameRules::ClientDisconnected( edict_t *pClient )
{
// [...]      
      // Tell all clients this player isn't a spectator anymore
      MESSAGE_BEGIN( MSG_ALL, gmsgSpectator );  
        WRITE_BYTE( ENTINDEX(pPlayer->edict()) );
        WRITE_BYTE( 0 );
      MESSAGE_END();
```

The NULL nicknames are taken care of with the current code, though, at least.

YaLTeR:
Alright, well, maybe no use in rebuilding teams in the condition then.

chinese-soup:
> Alright, well, maybe no use in rebuilding teams in the condition then.

It doesn't really do any harm to have it there, imho, since it means next time Draw() is called this if statement will no longer be met because NULL nicknames will be purged by then and therefore it won't have to skip the NULL nicknames anymore. Since unless you do this call they **will** be there while +showscores is held until the next RebuildTeams happens (which happens on someone model change/connect/disconnect/stop holding +showscores open scoreboard again/score change/etc.), so I'd personally just keep it there, I guess.

YaLTeR:
Let's go with this then.

PR #121, opened 2021-06-06
Add a CVAR for angled viewmodel bobbing

execut4ble:
Some players actually prefer the old style of viewmodel bobbing and the changes in #68 did not provide a way to toggle between them.


YaLTeR:
Maybe it shouldn't change the default behavior?

execut4ble:
The default behavior was changed with the previous changes, I intend to revert the default to how it was since I suspect this is more desirable than the angled viewbobbing.

YaLTeR:
I see, thanks.

PR #129, opened 2021-06-10
Sync up the condition for cl_righthand with the one that's in engine

### LINKED ISSUE #128: cl_righthand Bug
The command <cl_righthand> can be set to negative values, doing so inverts the view models inside out

chinese-soup:
Resolves #128 

Looks like an engine bug (couldn't really find where it happens in Ghidra, but it's pretty obvious seeing as we don't use cl_righthand's *actual* value anywhere in the client code), so not much we can do here other than clamp it.

* ~~Not just using min max (without if statement) since this way it will only be called once.~~
* ~~Resetting the cvar using Cvar_SetValue instead of min maxing the value, because engine is at fault and uses the cvar's value.~~

EDIT: alright yea looks like this is it

```c++
  (*qglCullFace)(0x404);
  bVar4 = false;
  if (((cl_righthand != (cvar_t *)0x0) && (0.0 < cl_righthand->value)) &&
     (cl.viewent.index == currententity->index)) {
    (*qglDisable)(0xb44);
    bVar4 = true;
  }
```

### DIFF
--- a/cl_dll/StudioModelRenderer.cpp
+++ b/cl_dll/StudioModelRenderer.cpp
@@ -983,7 +983,7 @@ void CStudioModelRenderer::StudioSetupBones ( void )
 			extern cvar_t* cl_righthand;
 			if (m_pCurrentEntity == gEngfuncs.GetViewModel()
 				&& IEngineStudio.IsHardware()
-				&& cl_righthand->value != 0.0f)
+				&& cl_righthand->value > 0.0f)
 			{
 				for (size_t j = 0; j < 4; ++j)
 					bonematrix[1][j] *= -1.0;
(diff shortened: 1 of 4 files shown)


tmp64:
The engine checks whether `cl_righthand` is enabled using `value > 0.0` whicle the client uses `value != 0.0`. This somehow inverts back surface culling.
```cpp
void R_GLStudioDrawPoints() {
  ...
  if ( cl_righthand && cl_righthand->value > 0.0 && ei.viewent.index == currententity->index )
  {
    qglDisable(0xB44u);
    iFlippedVModelb = 1;
  }
```

I think a better fix would be to replace all `!= 0` conditions with `> 0` like in the engine.

chinese-soup:
Oh yea, I found that code as well, but I didn't read the if statement correctly before so it didn't occur to me at all. :smile: 

> I think a better fix would be to replace all `!= 0` conditions with `> 0` like in the engine.
Good point, did this instead, thanks!

YaLTeR:
Thanks

PR #131, opened 2021-06-13
Rainbow HUD: Fix negative phase

tmp64:
If the game has just started (`gHUD.m_flTime` is low) and x or y are negative, current phase angle in `GetRainbowColor` can become negative. Then `fmod 360`, which returns values in the range (-360; 360), can return a negative number and trigger the assetion in `HSVtoRGB`.

This PR resolves that by adding 360 degrees to fmod result if it's negative.


YaLTeR:
Thanks

PR #142, opened 2021-08-11
Always render local player model when watching a POV demo with dem_forcehltv 1

chinese-soup:
Hello. Tried to make the cvar name as close to the dem_forcehltv since it only applies to this (because in a real HLTV demo, the localplayer is the HLTV proxy and therefore all players are rendered correctly) and at the same time not too long.
This is basically this, but client DLL edition: https://github.com/YaLTeR/bxt-rs/issues/10 (https://github.com/advancedfx/advancedfx/blob/main/AfxHookGoldSrc/HltvFix.cpp)

Video:
https://user-images.githubusercontent.com/5108747/129068510-f3c7a689-5bce-4e59-bcf7-3fd655195948.mp4

### DIFF
--- a/cl_dll/in_camera.cpp
+++ b/cl_dll/in_camera.cpp
@@ -18,6 +18,7 @@
 #include "pm_shared.h"
 
 #include "port.h"
+#include "demo_api.h"
 
 float CL_KeyState (kbutton_t *key);
 
@@ -64,7 +65,6 @@ cvar_t	*c_maxyaw;
 cvar_t	*c_minyaw;
 cvar_t	*c_maxdistance;
 cvar_t	*c_mindistance;
-cvar_t	*dem_forcehltv_local_model;
 
 // pitch, yaw, dist
 vec3_t cam_ofs;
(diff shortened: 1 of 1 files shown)


YaLTeR (commented):
Maybe it should just be always on, without a cvar? Is there any reason to ever have it off?

chinese-soup:
> Maybe it should just be always on, without a cvar? Is there any reason to ever have it off?

Looked into it and couldn't find any "wrong thirdperson" when enabling this, therefore I've changed it to be always on for when dem_forcehltv is set to 1 (which is the same as being connected to a HLTV server / watching a real HLTV demo) and demo playback is happening, which shouldn't do any regression for POV demos, HLTV demos or live HLTV watching. (I've checked and the HLTV proxy "player" isn't rendered accidentally or anything).

cl_dll/in_camera.cpp
```diff
@@ -15,8 +15,10 @@
 #include "camera.h"
 #include "in_defs.h"
 #include "Exports.h"
+#include "pm_shared.h"
 
 #include "port.h"
+#include "demo_api.h"
```
YaLTeR, line 21: both headers needed?

chinese-soup: yep, one for obs server mode macros, one for pDemoAPI struct

YaLTeR:
Thanks!

Issue #53, opened 2017-12-09
HLKZ quality of life features


YaLTeR:
- [ ] Automatic demo stop-record on teleport to start position / respawn
- [ ] Automatic demo stop-rename on pressing the finish button, configurable to continue recording for up to n seconds after the button press (or up to the next start position / respawn (configurable to only with PB?)

execut4ble:
- [ ] Add ability to create clientside checkpoints that work like custom BXT triggers. It should be able to detect when you start the timer and once you pass through the checkpoint it will show how much you're behind/ahead your previous attempts. Would also be cool if you could export/save checkpoints and times into a file like you can with splits.

- [ ] Ability to hide other players. Probably fine if they still make sounds but are invisible, if it's impossible to remove that.

- [ ] HLKZ/Gamemode configs. I use slightly different binds on HLKZ than when I'm playing regular AG so it would be nice if I could point a certain config I want to use when playing HLKZ and then I don't have to rebind stuff all the time.

execut4ble:
Invisible players have been implemented serverside in the HL KreedZ plugin.

PR #149, opened 2021-11-26
Strafing HUD

Schlufi:
strafe hud demo
This was suggested before in #53. Basically, the center of the bar tells you whether you're accelerating (green), decelerating (red), or neither (white). Note that this doesn't take into account ground friction (yet).
`hud_strafeguide 1` to enable. `hud_strafeguide_fov`, `hud_strafeguide_height`, `hud_strafeguide_size` to manipulate.
I don't think this is anywhere near as useful as the Defrag HUDs people use, unless you're a strafing god you'll likely be almost always stuck at the edge between green and white. Suggestions on how to improve this are welcome, this is more of a prototype than anything else. I might add an indicator for the region of maximum acceleration, that could make it more useful at least for circle jumping.
But at least it looks cool.

### DIFF
--- a/cl_dll/hud_strafeguide.cpp
+++ b/cl_dll/hud_strafeguide.cpp
@@ -1,3 +1,5 @@
+#define _USE_MATH_DEFINES
+#include <cmath>
 #include <complex>
 
 #include "hud.h"
@@ -75,7 +77,7 @@ int CHudStrafeGuide::Draw(float time)
 			boxLeft  *= zoom;
 			boxRight *= zoom;
 			
-			if (std::abs(boxLeft) > fov && std::abs(boxRight) > fov)
+			if (std::abs(boxLeft) > fov && std::abs(boxRight) > fov && boxRight * boxLeft > 0)
 				continue;
 			
 			boxLeft  = boxLeft  > fov ? fov : boxLeft  < -fov ? -fov : boxLeft;
--- a/cl_dll/hud_strafeguide.cpp
+++ b/cl_dll/hud_strafeguide.cpp
@@ -13,14 +13,12 @@ enum border {
 	GREEN_RED
 };
 
-const float pi = std::acos(-1.);
-
 int CHudStrafeGuide::Init()
 {
 	m_iFlags = HUD_ACTIVE;
 
 	hud_strafeguide = CVAR_CREATE("hud_strafeguide", "0", FCVAR_ARCHIVE);
-	hud_strafeguide_fov = CVAR_CREATE("hud_strafeguide_fov", "140", FCVAR_ARCHIVE);
+	hud_strafeguide_zoom = CVAR_CREATE("hud_strafeguide_zoom", "1", FCVAR_ARCHIVE);
 	hud_strafeguide_height = CVAR_CREATE("hud_strafeguide_height", "0", FCVAR_ARCHIVE);
 	hud_strafeguide_size = CVAR_CREATE("hud_strafeguide_size", "0", FCVAR_ARCHIVE);
 
@@ -38,7 +36,8 @@ int CHudStrafeGuide::Draw(float time)
 	if (hud_strafeguide->value == 0)
 		return 0;
 	
-	double fov = hud_strafeguide_fov->value / 2 / 180 * pi;
+	double fov = gHUD.default_fov->value / 2 / 180 * M_PI;
+	fov /= hud_strafeguide_zoom->value;
 	
 	int size = gHUD.m_iFontHeight;
 	int height = ScreenHeight / 2 - 2*size;
@@ -93,9 +92,9 @@ int CHudStrafeGuide::Draw(float time)
 
 static double angleReduce(double a)
 {
-	double tmp = std::fmod(a, 2*pi);
-	if (tmp < 0) tmp += 2*pi;
-	if (tmp > M_PI) tmp -= 2*pi;
+	double tmp = std::fmod(a, 2*M_PI);
+	if (tmp < 0) tmp += 2*M_PI;
+	if (tmp > M_PI) tmp -= 2*M_PI;
 	return tmp;
 }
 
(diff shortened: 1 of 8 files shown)


cl_dll/hud_strafeguide.cpp
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const float pi = std::acos(-1.);
+
+int CHudStrafeGuide::Init()
+{
+	m_iFlags = HUD_ACTIVE;
+
+	hud_strafeguide = CVAR_CREATE("hud_strafeguide", "0", FCVAR_ARCHIVE);
+	hud_strafeguide_fov = CVAR_CREATE("hud_strafeguide_fov", "140", FCVAR_ARCHIVE);
```
YaLTeR, line 23: Shouldn't it just use the current FOV?

cl_dll/hud_strafeguide.cpp
```diff
@@ -0,0 +1,157 @@
...
+	hud_strafeguide_height = CVAR_CREATE("hud_strafeguide_height", "0", FCVAR_ARCHIVE);
+	hud_strafeguide_size = CVAR_CREATE("hud_strafeguide_size", "0", FCVAR_ARCHIVE);
```
YaLTeR, line 25: What does this do? Pretty confusing to have both height and size for a full-screen bar.

cl_dll/hud_strafeguide.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include "pm_movevars.h"
+
+enum border {
+	RED_GREEN,
+	GREEN_WHITE,
+	WHITE_GREEN,
+	GREEN_RED
+};
+
+const float pi = std::acos(-1.);
```
YaLTeR, line 16: M_PI?

cl_dll/hud_strafeguide.cpp
```diff
@@ -0,0 +1,157 @@
...
+	double tmp = std::fmod(a, 2*pi);
+	if (tmp < 0) tmp += 2*pi;
+	if (tmp > M_PI) tmp -= 2*pi;
+	return tmp;
+}
+
+void CHudStrafeGuide::Update(struct ref_params_s *pparams)
+{
+	double frameTime = pparams->frametime;
+	auto input = std::complex<double>(pparams->cmd->forwardmove, pparams->cmd->sidemove);
```
YaLTeR, line 105: A bit cursed to use complex for vectors but I guess it works

YaLTeR (commented):
This is pretty interesting! I'll have to try it out myself.

I think a 1 px wide optimal strafing angle indicator would be cool, e.g. with bright green.

cl_dll/hud_strafeguide.cpp
tmp64: There is `Vector2D` in the SDK with math operators, I'd use it instead
https://github.com/YaLTeR/OpenAG/blob/master/cl_dll/util_vector.h

cl_dll/hud_strafeguide.cpp
Schlufi: I thought that people might want to change it. It controls the overall horizontal scale, and can be quite small.You can see in the image in the PR that at 300-ish speeds the green region is already pretty tiny with my FOV of 140.
Looking at it again, it's obvious that `hud_strafeguide_fov 0` should refer to using `default_fov`, and should be the default choice.
I know you don't like having lots of cvars, so I thought I'd start by putting in every one I could think of and then letting you decide which ones you want to keep :^) It's fine if you want to get rid of this one.

YaLTeR: I'll need to check how it looks myself in-game to see how much this "zooming in" is needed. Speaking of, if this is indeed for "zooming in" then perhaps it should be called zoom and have values like 1, 1.5, 2, etc?

cl_dll/hud_strafeguide.cpp
Schlufi: `hud_strafeguide_height` refers to the vertical position to be consistent with `hud_speedometer_height` and `hud_jumpspeed_height`. That's why the naming is so awkward, `hud_strafeguide_size` controls the fatness of the bar.
At first I made it always use the font size for the size of the bar. That looks good at high (4K) resolution, but at lower resolutions I think it's way too big.

cl_dll/hud_strafeguide.cpp
Schlufi: Yeah that's just a mistake.

cl_dll/hud_strafeguide.cpp
Schlufi: `Vector2D` was the first thing I used, but I couldn't make the  `#include`s work out right to actually use it properly. Tracing it through, I think `hud.h` (which indirectly brings in `util_vector.h`) needs to be included before `cl_util.h`,  but `cl_util.h` defines a macro for `DotProduct` which is incompatible with the Vector2D function of the same name...
I'm not really a C++ programmer so at that point I decided to go with the simpler solution of complex numbers. Also makes it easier to get the angle of a vector because you can just use a complex logarithm lol.

Schlufi:
Here's a quick demo of what it looks like right now: https://files.catbox.moe/t0v1vw.webm
It's pretty twitchy when you switch directions a lot, but I guess that's just how this game is.

cl_dll/hud_strafeguide.cpp
```diff
@@ -120,24 +119,30 @@ void CHudStrafeGuide::Update(struct ref_params_s *pparams)
 
 	bool onground = pparams->onground;
 	double accelCoeff = onground ? pparams->movevars->accelerate : pparams->movevars->airaccelerate;
-	double frictionCoeff = pparams->movevars->friction;
-	double speedCap = onground ? pparams->movevars->maxspeed : min(pparams->movevars->maxspeed, 30);
+	//TODO: grab the entity friction from somewhere. pparams->movevars->friction is sv_friction
```
dMolite, line 122: when i am still not mistaken: playermove_s stores alteration from gravity / friction or conveyor belts

YaLTeR:
I finally gave it a try (sorry for taking so long!). This is really neat! I just have a question about zoom: it feels like zoom 1 doesn't actually correspond to the camera view, as I would expect. I.e., at zoom 1 I expect the green bar turn to red at the point on the screen where I would need to aim this frame to go into the red area, while in-game it feels like the bars are too big for that (but maybe I'm mistaken).

YaLTeR:
I suppose this is fine now then, just need to fix the CI failure?

YaLTeR:
Thanks

PR #150, opened 2021-11-29
Add cvars to ignore certain messages

fireblizzard:
With the next version of AG an extra byte will be sent on the TextMsg event. This byte tells what type of message it is. For the moment I have added 2 message types, and I'm planning a few more in the future, like a type for name and/or team changing messages. This makes it possible to ignore certain types of messages easily in the client, without relying on server configuration or checking what kind of text the message has.

This byte will be:
-1: if absent (that's what READ_BYTE() returns if it ran out of bytes to read)
0: if it's a spawn message
1: if it's a damage message

A spawn message is sent as soon as a practice bot respawns, telling the spot number where it spawned.
A damage message tells how much damage you dealt to a practice bot.

New cvars:

- `cl_ignore_spawn_messages` <0/1> - Default: 0 (don't ignore this message type)
- `cl_ignore_damage_messages` <0/1> - Default: 0 (don't ignore this message type)

As I mentioned there will be more types, for name/team changes, as some players have requested the ability to ignore those because some spectators like to change their name/model to distract players during a match, or to give unfair advantage to one of them telling the opponent's location.

Any suggestion is very welcome, regarding anything really (cvar names, coding, ignoring strategy, etc.). My C++ knowledge is very limited as you know :)

### DIFF
--- a/cl_dll/hud.h
+++ b/cl_dll/hud.h
@@ -418,6 +418,8 @@ struct message_parms_t
 
 class CHudTextMessage: public CHudBase
 {
+	std::vector<cvar_t*> ignored_message_types;
+
 public:
 	int Init( void );
 	static char *LocaliseTextString( const char *msg, char *dst_buffer, int buffer_size );
--- a/cl_dll/text_message.cpp
+++ b/cl_dll/text_message.cpp
@@ -34,6 +34,9 @@ int CHudTextMessage::Init(void)
 {
 	HOOK_MESSAGE( TextMsg );
 
+	ignored_message_types.push_back(CVAR_CREATE("cl_ignore_spawn_messages", "0", FCVAR_ARCHIVE));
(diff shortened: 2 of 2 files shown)


YaLTeR (approved):
Seems good; I'm assuming it doesn't break compatibility with vanilla AG clients and servers?

fireblizzard:
Yeah, I tested these combinations, they are all good:

- Vanilla 6.6 client on a latest 6.7 server
- Latest OpenAG release (Jun 3 2021) on a latest 6.7 server
- This PR version on a latest 6.7 server, and on a 6.7 listenserver too
- This PR version on a vanilla 6.6 server
- This PR version on a server with an older 6.7 build with HLKZ

The test was just triggering different messages by saying random stuff, changing model, changing name, killing several times a bot that shows the spawn and damage messages, and replaying demos on HLKZ. Also changing the implemented cvar values to ignore one message type, then the other, then both; it worked fine.

YaLTeR:
Sounds great, thanks!

PR #155, opened 2022-02-20
Ducktap fix

yhoiluhi:
Fix +ducktap that you weren't able to interfere in "ducking" process, now its working like +bxt_tas_ducktap

### DIFF
--- a/cl_dll/input.cpp
+++ b/cl_dll/input.cpp
@@ -162,22 +162,23 @@ namespace autofuncs
 	static void handle_ducktap(usercmd_t* cmd)
 	{
 		static bool s_duck_was_down_last_frame = false;
-		static bool duck_was_pressed = false;
 		static bool should_release_duck;
 
-		if (in_duck.state & 1)
-			duck_was_pressed = true;
-		else
-			duck_was_pressed = false;
+		bool duck_is_pressed = false;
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
Is there any reason why you moved the code down? It's very hard to see what the actual code differences are this way. Could you move it back where it was so the difference is readable please?

yhoiluhi:
I'm using in_duck.state check in function which is undeclared if i don't move function down

cl_dll/input.cpp
```diff
@@ -182,6 +125,74 @@ typedef struct kblist_s
...
+		s_jump_was_down_last_frame = ((cmd->buttons & IN_JUMP) != 0);
+	}
+
+	static void handle_ducktap(usercmd_t* cmd)
+	{
+		static bool s_duck_was_down_last_frame = false;
+		static bool duck_was_pressed = false;
+		static bool should_release_duck;
+
+		if (in_duck.state & 1)
```
YaLTeR, line 168:
```suggestion
		static bool should_release_duck;

		bool duck_was_pressed = false;
		if (in_duck.state & 1)
```
Doesn't need to be a static if it's set every time.

Also please rename to `duck_is_pressed` and remove the set in the `should_release_duck` condition below as it does nothing.

YaLTeR:
Thanks!

PR #170, opened 2022-08-16
SayText: Add cvar to disable rainbow color

### LINKED ISSUE #169: Make hud_rainbow 1 no to affect the chat
Well, you might have heard this before, but making the chat messages "rainbow" it's kind of annoying since you can't differenciate colors in your teammates' messages, locs and so. I know that in "bugfixed" the rainbow hud doesn't affect the chat, but I was wondering if it can be donde in OpenAG as well, maybe as toggle because there might be people who likes the rainbow chat.

tmp64:
Resolves #169


cl_dll/saytext.cpp
```diff
@@ -105,6 +106,10 @@ int CHudSayText :: Draw( float flTime )
 	if ( ( gViewPort && gViewPort->AllowedToPrintText() == FALSE) || !m_HUD_saytext->value )
 		return 1;
 
+	bool enableRainbow = m_HUD_rainbow_chat->value;
+	if (!enableRainbow)
+		gHUD.m_Rainbow.PushDisable();
```
YaLTeR, line 111: Is this recursive? I.e. why does this need a counter instead of a simple true/false?

tmp64: Futureproofing. With a true/false code like this will unexpectedly break:
```cpp
void SomeCommonFunc() {
  SetEnabled(false);
  // draw something
  SetEnabled(true);
}

void DrawSomeElement() {
  SetEnabled(false);
  // ...
  SomeCommonFunc();
  // Effect now enabled
  // ...
  SetEnabled(true);
}
```

YaLTeR:
Thanks!

PR #171, opened 2022-10-03
CustomTimer: Use float time instead of decimal

SmileyAG:


SmileyAG:
> Why?

Useful when I want to measure with checkpoints for how much some route differences in HLKZ with a bind like:

`bind v "stop;record route;say /practicetp;customtimer 20" // teleport to latest practice checkpoint`

cl_dll/hud_customtimer.cpp
```diff
@@ -34,7 +34,7 @@ int CHudCustomTimer::Draw(float time)
 	UnpackRGB(r, g, b, gHUD.m_iDefaultHUDColor);
 
 	FillRGBA((ScreenWidth / 2) - 50, (gHUD.m_scrinfo.iCharHeight * 4) - 6, (diff / m_flSeconds) * 100, 4, r, g, b, 210);
-	snprintf(str, sizeof(str), "Timer: %d", (int)diff);
+	snprintf(str, sizeof(str), "Timer: %.1f", diff);
```
YaLTeR, line 37:
```suggestion
	snprintf(str, sizeof(str), "Timer: %.01f", diff);
```
I think this is needed for it to show .0

SmileyAG: It does show even without that change.

YaLTeR:
Thanks

PR #172, opened 2022-10-12
Added cl_hide_other_players

chinese-soup:
image

(Some additional info on his use case is: 1) doesnt hide players on HLKZ when playin', 2) recorded POV demo, 3) got a good run, 4) wanted to render POV demo, but now the other player is visible on the demo 5) sadface 6) this pr)

* No FCVAR_ARCHIVE so ppl dont forget that they hid players and wonder "wtf where players at"


YaLTeR:
Thanks

PR #174, opened 2022-12-09
Make hud_saytext draw even when hud_draw is 0

SmileyAG:
Can be useful for recording some HLKZ runs (because there is always chat output of run time/top in leaderboard)


YaLTeR:
Not sure showing the chat output is in the spirit of hud_draw 0... Also not sure how to go about this sort of request. What are the use-cases for hud_draw 0? Do people use it during gameplay, or just for demo recording? There's HLKZ demo recording (not necessarily with chat output, you might very reasonably want to overlay the time afterwards in a video editor in a nice way), there's also frag movie demo recording (where you likely don't want any chat output), anything else?

YaLTeR:
Apparently Bugfixed HL also draws it with hud_draw 0, so that's good then.

PR #176, opened 2022-12-24
Make name/team change and game leave messages ignorable

fireblizzard:
AG 6.7 sends a byte for these chat message types that we can use to ignore them


YaLTeR:
Thanks

PR #183, opened 2023-12-06
Added cl_show_server_triggers_* cvars (related to plugin PR)

SmileyAG:
Explained in https://github.com/YaLTeR/hlkreedz/pull/37

I recently remembered and rebased my old request for plugin a few days ago and at the same time reminded about it to naz in discord, he said that would approve later, so I’m sending a pull request in advance, it’s not necessary to merge it now (since the PR plugin isn't merged yet), but you shouldn’t close it either

### DIFF
--- a/cl_dll/cl_util.h
+++ b/cl_dll/cl_util.h
@@ -121,8 +121,6 @@ static size_t count_digits(int n)
 
 static size_t get_map_name(char* dest, size_t count)
 {
-	memset(dest, 0, count);
-
 	auto map_path = gEngfuncs.pfnGetLevelName();
 
 	const char* slash = strrchr(map_path, '/');
--- a/cl_dll/cl_util.h
+++ b/cl_dll/cl_util.h
@@ -121,6 +121,8 @@ static size_t count_digits(int n)
 
 static size_t get_map_name(char* dest, size_t count)
 {
+	memset(dest, 0, count);
+
 	auto map_path = gEngfuncs.pfnGetLevelName();
 
 	const char* slash = strrchr(map_path, '/');
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -1,5 +1,6 @@
 // Client side entity management functions
 
+#include <algorithm>
 #include <memory.h>
 
 #include "hud.h"
@@ -19,6 +20,10 @@
 #include "discord_integration.h"
 
 #include "particleman.h"
+
+#undef min
+#undef max
+
 extern IParticleMan *g_pParticleMan;
 
 void Game_AddObjects( void );
@@ -49,6 +54,11 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 	default:
 		break;
 	}
+
+	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
+	const int kRenderFxTrigger = 241;
+	if (ent->curstate.rendermode == kRenderTransColor && ent->curstate.renderfx == kRenderFxTrigger && gHUD.m_pShowServerTriggers->value > 0)
+		ent->curstate.renderamt = std::min(255.0f, std::max(0.0f, gHUD.m_pShowServerTriggersAlpha->value));
 
 	// hide corpses option
 	if (gHUD.m_pCvarHideCorpses->value > 0 && ent->curstate.renderfx == kRenderFxDeadPlayer)
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -21,6 +21,10 @@
 
 #include "particleman.h"
 
+#include "r_studioint.h"
+
+extern engine_studio_api_t IEngineStudio;
+
 #undef min
 #undef max
 
@@ -57,8 +61,8 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 
 	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
 	// update: there is a new implementation of displaying triggers that allows you to display even when planes is stripped due to optimizations in updated map compiler
-	// so this code will only work if the value 2 is specified in the cvar, but it should not be deleted imo
-	if ((ent->curstate.rendermode == kRenderTransColor) && (ent->curstate.renderfx == kRenderFxTrigger) && (gHUD.m_pShowServerTriggers->value == 2.0f) && !gHUD.IsTriggerForSinglePlayer(ent->curstate.rendercolor))
+	// so for hardware engine this code will only work if the value 2 is specified in the cvar
+	if ((ent->curstate.rendermode == kRenderTransColor) && (ent->curstate.renderfx == kRenderFxTrigger) && (gHUD.m_pShowServerTriggers->value == 2.0f || (!IEngineStudio.IsHardware && gHUD.m_pShowServerTriggers->value > 0)) && !gHUD.IsTriggerForSinglePlayer(ent->curstate.rendercolor))
 		ent->curstate.renderamt = std::min(255.0f, std::max(0.0f, gHUD.m_pShowServerTriggersAlpha->value));
 
 	// hide corpses option
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -32,6 +32,26 @@ extern vec3_t v_origin;
 
 int g_iAlive = 1;
 
+bool IsTriggerForSinglePlayer(color24 rendercolor)
+{
+	auto r = rendercolor.r;
+	auto g = rendercolor.g;
+	auto b = rendercolor.b;
+
+	if ((r == 128) && (g == 128) && (b == 128)) // trigger_autosave
+		return true;
+	else if ((r == 79) && (g == 255) && (b == 10)) // trigger_changelevel
+		return true;
+	else if ((r == 150) && (g == 75) && (b == 0)) // trigger_endsection
+		return true;
+	else if ((r == 238) && (g == 154) && (b == 77)) // trigger_monsterjump
+		return true;
+	else if ((r == 203) && (g == 103) && (b == 212)) // trigger_transition
+		return true;
+
+	return false;
+}
+
 /*
 ========================
 HUD_AddEntity
@@ -57,7 +77,7 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 
 	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
 	const int kRenderFxTrigger = 241;
-	if (ent->curstate.rendermode == kRenderTransColor && ent->curstate.renderfx == kRenderFxTrigger && gHUD.m_pShowServerTriggers->value > 0)
+	if ((ent->curstate.rendermode == kRenderTransColor) && (ent->curstate.renderfx == kRenderFxTrigger) && (gHUD.m_pShowServerTriggers->value > 0) && !IsTriggerForSinglePlayer(ent->curstate.rendercolor))
 		ent->curstate.renderamt = std::min(255.0f, std::max(0.0f, gHUD.m_pShowServerTriggersAlpha->value));
 
 	// hide corpses option
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -23,8 +23,6 @@
 
 #include "r_studioint.h"
 
-extern engine_studio_api_t IEngineStudio;
-
 #undef min
 #undef max
 
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -32,26 +32,6 @@ extern vec3_t v_origin;
 
 int g_iAlive = 1;
 
-bool IsTriggerForSinglePlayer(color24 rendercolor)
-{
-	auto r = rendercolor.r;
-	auto g = rendercolor.g;
-	auto b = rendercolor.b;
-
-	if ((r == 128) && (g == 128) && (b == 128)) // trigger_autosave
-		return true;
-	else if ((r == 79) && (g == 255) && (b == 10)) // trigger_changelevel
-		return true;
-	else if ((r == 150) && (g == 75) && (b == 0)) // trigger_endsection
-		return true;
-	else if ((r == 238) && (g == 154) && (b == 77)) // trigger_monsterjump
-		return true;
-	else if ((r == 203) && (g == 103) && (b == 212)) // trigger_transition
-		return true;
-
-	return false;
-}
-
 /*
 ========================
 HUD_AddEntity
@@ -76,8 +56,9 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 	}
 
 	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
-	const int kRenderFxTrigger = 241;
-	if ((ent->curstate.rendermode == kRenderTransColor) && (ent->curstate.renderfx == kRenderFxTrigger) && (gHUD.m_pShowServerTriggers->value > 0) && !IsTriggerForSinglePlayer(ent->curstate.rendercolor))
+	// update: there is a new implementation of displaying triggers that allows you to display even when planes is stripped due to optimizations in updated map compiler
+	// so this code will only work if the value 2 is specified in the cvar, but it should not be deleted imo
+	if ((ent->curstate.rendermode == kRenderTransColor) && (ent->curstate.renderfx == kRenderFxTrigger) && (gHUD.m_pShowServerTriggers->value == 2.0f) && !gHUD.IsTriggerForSinglePlayer(ent->curstate.rendercolor))
 		ent->curstate.renderamt = std::min(255.0f, std::max(0.0f, gHUD.m_pShowServerTriggersAlpha->value));
 
 	// hide corpses option
--- a/cl_dll/entity.cpp
+++ b/cl_dll/entity.cpp
@@ -61,8 +61,8 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 
 	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
 	// update: there is a new implementation of displaying triggers that allows you to display even when planes is stripped due to optimizations in updated map compiler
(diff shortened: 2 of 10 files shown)


tmp64:
Maybe check for `gEngfuncs.GetMaxClients() > 1` when hiding singleplayer triggers?

SmileyAG:
> Maybe check for `gEngfuncs.GetMaxClients() > 1` when hiding singleplayer triggers?

Well, I'd say it might require a separate cvar at that point then, since even when I'm playing at own listenserver with `maxplayers 1` and loaded amxmodx+metamod-p+hlkreedz combo, still don't want to see those SP triggers because they wouldn't action normally in deathmatch mode, example: https://github.com/YaLTeR/OpenAG/blob/27e7143daee4823b12cda4bcd1f1e8d1fbdaa4d6/dlls/triggers.cpp#L1466-L1475

SmileyAG:
Please review this code again, because now I have fully changed it to display triggers with TriAPI

Let me first explain why this was done, because if being unenlightened in that question, then you would have a fair reason to ask: why the hell was it needed to go this way?

In short, modern map compilers cut off unused planes due to optimization (such as from triggers), but trigger entities from plugin is still transmitted from the server to the client, which means we can try to take their `mins`, `maxs`, `origin` and draw them separately (e.g. with TriAPI)

Showcase (value 1 = old triggers, value 2 = new triggers):

https://github.com/YaLTeR/OpenAG/assets/58108407/dd59c2b3-130c-43d6-a55e-c83498221762

tmp64:
What about trigger brushes of forms other than a box? They would be rendered incorrectly

cl_dll/tri.cpp
```diff
@@ -1,4 +1,4 @@
-//========= Copyright � 1996-2002, Valve LLC, All rights reserved. ============
+//========= Copyright � 1996-2002, Valve LLC, All rights reserved. ============
```
tmp64, line 1: Broken character became broken slightly differently

cl_dll/tri.cpp
```diff
@@ -40,6 +59,133 @@ void CL_DLLEXPORT HUD_DrawNormalTriangles( void )
...
+	pTriAPI->Vertex3f(corner1.x, corner2.y, corner1.z);
+	pTriAPI->Vertex3f(corner1.x, corner2.y, corner2.z);
+
+	pTriAPI->End();
+}
+
+static std::vector<Vector> trigger_mins;
+static std::vector<Vector> trigger_maxs;
+static std::vector<Vector> trigger_origin;
+static std::vector<color24> trigger_rgb;
```
tmp64, line 110: Why not make a vector of structs?

cl_dll/tri.cpp
```diff
@@ -57,4 +203,28 @@ void CL_DLLEXPORT HUD_DrawTransparentTriangles( void )
 
 	if ( g_pParticleMan )
 		 g_pParticleMan->Update();
+
+	// Draw server-side triggers with TriAPI instead
+
+	glDisable(GL_TEXTURE_2D);
```
tmp64, line 209: Does this work in software mode?

cl_dll/tri.cpp
tmp64: It was broken on the review page but it's visible in the comments...

cl_dll/tri.cpp
SmileyAG: Haven't tested but probably not, so yeah I could check for `IEngineStudio.IsHardware()` at that point then

SmileyAG:
> What about trigger brushes of forms other than a box? They would be rendered incorrectly

Probably, but have you ever seen a map in which the trigger is not a box? I've looked before through a lot of maps in the `bspguy` or directly in map sources with `J.A.C.K`, and I've never seen a brush trigger that wasn't in a box shape

cl_dll/tri.cpp
SmileyAG: Even when you are completely sure that auto-formatting is disabled and encoding should not be changed, IDE goes brrrrrrrrrrrrrrrrrr

Thanks for noting actually, would to fix then

tmp64:
> Probably, but have you ever seen a map in which the trigger is not a box? I've looked before through a lot of maps in the `bspguy` or directly in map sources with `J.A.C.K`, and I've never seen a brush trigger that wasn't in a box shape

You can try rendering the 0-th hull of the brush entity model. It should be the exact one used for collision detection with point entities.  
https://github.com/dreamstalker/rehlds/blob/93f5775ac26240782981f47ee8e052fb53d30877/rehlds/engine/world.cpp#L193

YaLTeR:
Note that I'm not sure if we can use GPL code in OpenAG.

YaLTeR:
I think rendering boxes only is fine if the alternative complicates things a lot.

tmp64:
> Note that I'm not sure if we can use GPL code in OpenAG.

It's a reference to collision detection code and it's not intended to be added to OpenAG

cl_dll/tri.cpp
SmileyAG: Done

cl_dll/tri.cpp
SmileyAG: Done

cl_dll/tri.cpp
SmileyAG: Fixed

cl_dll/tri.cpp
SmileyAG: Oh, in fact for software support, it possible to replace `glDisable(GL_TEXTURE_2D)` / `glEnable(GL_TEXTURE_2D)` hack with:

```cpp
	static bool init = true;
	static HSPRITE white_sprite = 0;

	if (init)
	{
		white_sprite = gEngfuncs.pfnSPR_Load("sprites/white.spr");
		init = false;
	}

	if (white_sprite == 0)
		return;

	if (gEngfuncs.pTriAPI->SpriteTexture(const_cast<model_s*>(gEngfuncs.GetSpritePointer(white_sprite)), 0))
	{
		// Custom code
	}
```

cl_dll/cl_util.h
```diff
@@ -121,6 +121,8 @@ static size_t count_digits(int n)
 
 static size_t get_map_name(char* dest, size_t count)
 {
+	memset(dest, 0, count);
```
YaLTeR, line 124: This is not needed. This function always caps `dest` off with a zero.

cl_dll/entity.cpp
```diff
@@ -50,6 +59,12 @@ int CL_DLLEXPORT HUD_AddEntity( int type, struct cl_entity_s *ent, const char *m
 		break;
 	}
 
+	// show triggers that would be transferred from server-side with specific value in renderfx to differ it from other entities
+	// update: there is a new implementation of displaying triggers that allows you to display even when planes is stripped due to optimizations in updated map compiler
+	// so this code will only work if the value 2 is specified in the cvar, but it should not be deleted imo
```
YaLTeR, line 64: > but it should not be deleted imo

Why not? Is it preserved server-side too?

cl_dll/hud.cpp
```diff
@@ -937,3 +939,22 @@ float CHud::GetSensitivity( void )
 	return m_flMouseSensitivity;
 }
 
+bool CHud::IsTriggerForSinglePlayer(color24 rendercolor)
+{
+	auto r = rendercolor.r;
+	auto g = rendercolor.g;
+	auto b = rendercolor.b;
+
+	if ((r == 128) && (g == 128) && (b == 128)) // trigger_autosave
```
YaLTeR, line 948: It's pretty brittle comparing by color values like this, because now you're locked out of easily changing the colors if needed (need to change on the server + here). Is it possible to compare the classname instead?

cl_dll/hudgl.cpp
```diff
@@ -6,7 +6,9 @@
 #include <vector>
 
 #ifdef _WIN32
+#include <winsani_in.h>
```
YaLTeR, line 9: What's this for?

cl_dll/tri.cpp
```diff
@@ -40,6 +60,91 @@ void CL_DLLEXPORT HUD_DrawNormalTriangles( void )
...
+					{
+						color24 colors = ent->curstate.rendercolor;
+						if (!gHUD.IsTriggerForSinglePlayer(colors))
+						{
+							gEngfuncs.pTriAPI->RenderMode(kRenderTransAdd);
+							gEngfuncs.pTriAPI->CullFace(TRI_NONE);
+
+							float r = colors.r, g = colors.g, b = colors.b, a = std::min(255.0f, std::max(0.0f, gHUD.m_pShowServerTriggersAlpha->value));
+							DivideRGBABy255(r, g, b, a);
+							gEngfuncs.pTriAPI->Color4f(r, g, b, a);
```
YaLTeR, line 129: Maybe `Color4ub()` then you don't need to divide?

cl_dll/tri.cpp
```diff
@@ -57,4 +162,14 @@ void CL_DLLEXPORT HUD_DrawTransparentTriangles( void )
 
 	if ( g_pParticleMan )
 		 g_pParticleMan->Update();
+
+	if (gHUD.white_sprite == 0)
+		return;
+
+	if (gEngfuncs.pTriAPI->SpriteTexture(const_cast<model_s*>(gEngfuncs.GetSpritePointer(gHUD.white_sprite)), 0))
```
YaLTeR, line 169: `if (!...) return;`?

common/com_model.h
```diff
@@ -16,7 +16,7 @@
 #define STUDIO_EVENTS 2
 
 #define MAX_CLIENTS			32
-#define	MAX_EDICTS			900
+#define	MAX_EDICTS			2048
```
YaLTeR, line 19: Is this related?

cl_dll/tri.cpp
SmileyAG: Month ago, I had to reverse-engineering all TriAPI functions from hardware engine (since I am preparing for a major refactor update for BXT) and I can tell you that using `pTriAPI->Color4ub()` will not be a good idea, take a look yourself:

```cpp
void tri_GL_Color4ub(unsigned char r, unsigned char g, unsigned char b, unsigned char a) // pTriAPI->Color4ub
{
	gGlR = r * (1.0f / 255.0f);
	gGlG = g * (1.0f / 255.0f);
	gGlB = b * (1.0f / 255.0f);
	gGlW = a * (1.0f / 255.0f);
	glColor4f(gGlR, gGlG, gGlB, 1.0f);
}
```

Yeah, it divides all of that colors to 255, but as can be seen, `glColor4f` is would be still hardcoded to 1

If someone doubt my skills, well, open hardware linux engine in IDA Pro or Ghidra and you will see the same:

color4ub

YaLTeR: amazing

common/com_model.h
SmileyAG: Of course! That depending through how many entities we want to iterate to find triggers!

```cpp
if ((gHUD.m_pShowServerTriggers->value > 0) && (gHUD.m_pShowServerTriggers->value != 2.0f))
	{
		for (int e = 0; e < MAX_EDICTS; ++e)
		{
			cl_entity_t* ent = gEngfuncs.GetEntityByIndex(e);
			if (ent)
			{
```

According to the investigate from Solokiller, max edicts can be no more than 2047, anything higher than this will have cause crash: https://github.com/ValveSoftware/halflife/issues/1777

SmileyAG: I'd also started to think might be it should controllable instead? Like if the map has a small amonut of triggers, they you wouldn't do extra iterations, right?

Making some cvar like `cl_show_server_triggers_entities_limit 2048`

cl_dll/entity.cpp
SmileyAG: Eeeh, well, absmin/absmax triggers seems to be draw as should for the most of maps, but I had noticed that on some specific map (`dyd_axn_plant`) in a specific place it draw not the right size as should, so I'd keep that second mode for a reserve case, like if the main mode is would having some issues


Correct draw:

изображение


Wrong draw:

изображение

common/com_model.h
YaLTeR: CVar is definitely not the right solution here

YaLTeR: Let's just leave it as is for now, since it doesn't run with showtriggers 0

cl_dll/hud.cpp
SmileyAG: > It's pretty brittle comparing by color values like this, because now you're locked out of easily changing the colors if needed (need to change on the server + here). Is it possible to compare the classname instead?

There is only one method to get a classname on the client-side that I can think off, but believe me, this method will eat up performance even more and its implementation is a bit of a headache, but let me describe this concept to you briefly:

## First step
Get list of all entities on map by reading .bsp file
Or with `cl_entity_t *worldmodel = gEngfuncs.GetEntityByIndex(0);` and `worldmodel->model->entities`

## Second step

Let's take for example some entity from list for analyze:

```
{
"model" "*63"
"target" "tp_01"
"classname" "trigger_teleport"
}
```

In our case, we will be interested in only two keyvalues: `"classname"` and `"model"`

You may ask, why do we need the model name? Well, that is exactly the key to how we will differ them from other entities! Because the each brush entity has a unique model for it as mandatory!

# Third step

https://github.com/YaLTeR/OpenAG/blob/0ad666a1f2bee01da47e2880877d6d3c735be06b/cl_dll/hud_spectator.cpp#L237

Take the code from the `UTIL_FindEntityInMap` function as a basis and rewrite it to suit your needs, execute it only once (when changing the map) and write the results to the some `std::vector` variables

# Fourth step

```cpp
			for (int e = 0; e < MAX_EDICTS; ++e)
			{
				cl_entity_t* ent = gEngfuncs.GetEntityByIndex(e);
				if (ent && ent->model)
				{
					if (!entities_classname.empty() && !entities_model.empty())
					{
						for (size_t i = 0; i < entities_classname.size(); i++)
						{
							if (!strcmp(entities_model[i].c_str(), ent->model->name))
							{
								// Rest code
							}
						}
					}
				}
			}
```

SmileyAG: Therefore, I see the hardcoding colors on plugin side is rational way in that case, since there doesn't exist the straight way to get classname, only possible by the method that I described above

And besides, we can still change that colors on client-side, so that doesn't sound like issue at all

cl_dll/hudgl.cpp
SmileyAG: These headers are needed so that `ARRAYSIZE` and `HSPRITE` do not conflict with the HLSDK code, because Microsoft has its own implementation for them

Even if these changes are not directly related to this request, I see no reason not to add them, and I don’t want to create a separate request for that kind of stuff

cl_dll/tri.cpp
YaLTeR: well I actually meant the line that I commented on

SmileyAG: Honestly, I really don't like the concept of guard clauses when you have not much cases to check validity for on average

I usually prefer for code design to make them into nested if statements than exit from function each time when only x or y is not valid, because it more versality when for example you want later to add some code that should be executed exactly after that

Well, of course it still depends on case by case, but on average I would to prefer nested if statements instead

```cpp
// Pre

if (gHUD.white_sprite)
{
	// Might be there would be here some new code too, who knows?

	if (gEngfuncs.pTriAPI->SpriteTexture(const_cast<model_s*>(gEngfuncs.GetSpritePointer(gHUD.white_sprite)), 0))
	{
		// Triggers code
	}

	// And here too?
}

// Post
```

SmileyAG:
@YaLTeR @tmp64 Well I haven't contacted with naz for recent time, but at least exec approved that and updated plugin for all SourceRuns HLKZ servers, so pull requests in those both clients can be merged now:

изображение

cl_dll/tri.cpp
YaLTeR: Please make them both `if() return;`, that's what I was asking here. 🙃

> I usually prefer for code design to make them into nested if statements than exit from function each time when only x or y is not valid, because it more versatility when for example you want later to add some code that should be executed exactly after that

Then you turn them nested when you need it. Early return is better because it doesn't indent the code to the right, and because it's easier to follow the control flow. You don't need to keep track of the current conditions in the mental stack, because you consider them once, then they return, and you no longer need to think about them.

Designing for "something we might need to extend in the future" frequently leads to needlessly convoluted code, because right now you don't know what exactly you will need in the future. As a general principle, you should write code for the current needs only, and only in specific rare occasions design with extensibility from the get-go, while already having several use-cases in mind that you can design against.

SmileyAG: > Please make them both `if() return;`, that's what I was asking here. 🙃

I realized this even before I made the last commit, it’s just my preferences followed me to use nested if statements since there are not many cases that need to be checked for

But so to speak, your argument also makes sense that it is better to expand when you know it needed, like it good to make nested when you don't want to add some new code to it, but at the same time you have roadmap what in future you would exactly add here

cl_dll/hud.cpp
YaLTeR: Hmm, right. I don't see an obvious way to go from `cl_entity_t` to `entvars_t` (which has the classname).

cl_dll/tri.cpp
```diff
@@ -40,6 +60,91 @@ void CL_DLLEXPORT HUD_DrawNormalTriangles( void )
...
+							DrawAACuboid(gEngfuncs.pTriAPI, absmin, absmax);
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+extern engine_studio_api_t IEngineStudio;
```
YaLTeR, line 146: Is this used anywhere?

cl_dll/entity.cpp
```diff
@@ -19,6 +20,14 @@
 #include "discord_integration.h"
 
 #include "particleman.h"
+
+#include "r_studioint.h"
+
+extern engine_studio_api_t IEngineStudio;
```
YaLTeR, line 26: Is this used anywhere?

cl_dll/tri.cpp
SmileyAG: At first I used directly OpenGL functions and because of this I began to check whether we were using a hardware engine or not (`IEngineStudio.IsHardware()`), but later I abandoned this idea in favor of a solution that works optimally for both versions of engine

After that, I didn’t feel like to removing them, but I can do it you want

YaLTeR: Let's remove then this PR is good to go.

YaLTeR:
Thanks!

Issue #181, opened 2023-11-20
[LINUX] could not load library /home/.../ag/cl_dlls/client.so


Lamborge:
just install agmod and it won't start
gdb output:
`(gdb) call (void*) dlopen("ag/cl_dlls/client.so", 2)
$1 = (void *) 0x0`
`(gdb) call (char*) dlerror()
$2 = 0x8077690 "ag/cl_dlls/client.so: undefined symbol: SDL_SetRelativeMouseMode"`

YaLTeR:
Try with the "legacy" beta on Half-Life.

SmileyAG:
According to error, might be this has something to do that with the release of the 25th anniversary version, they removed the libSDL2 libs from the **Half-Life** folder for Linux depot: https://steamdb.info/depot/8/history/?changeid=M:2809540380682473923

Although the latest version of the engine is still dynamically linked to the libSDL2, it's just borrows version now from steam runtime (from `info sharedlibrary` in **gdb**): 

```
0xf5c6cda0  0xf5d8d978  Yes (*)     /home/smileyag/.steam/debian-installation/ubuntu12_32/steam-runtime/i386/usr/lib/i386-linux-gnu/libSDL2-2.0.so.0
```

Lamborge:
adding `LD_PRELOAD=/usr/lib32/libSDL2-2.0.so.0 %command%` to launch parameters solve the problem

SmileyAG:
> adding `LD_PRELOAD=/usr/lib32/libSDL2-2.0.so.0 %command%` to launch parameters solve the problem

Let me note that it is not necessary to use the system one `libSDL2-2.0.so.0`, you can also use it from steam runtime, the only important is that it must be preloaded:

```
LD_PRELOAD=$STEAM_RUNTIME/usr/lib/i386-linux-gnu/libSDL2-2.0.so.0 %command% -game ag
```

krangm:
> > adding `LD_PRELOAD=/usr/lib32/libSDL2-2.0.so.0 %command%` to launch parameters solve the problem
> 
> Let me note that it is not necessary to use the system one `libSDL2-2.0.so.0`, you can also use it from steam runtime, the only important is that it must be preloaded:
> 
> ```
> LD_PRELOAD=$STEAM_RUNTIME/usr/lib/i386-linux-gnu/libSDL2-2.0.so.0 %command% -game ag
> ```

[...]

PR #184, opened 2023-12-19
Make sure the client is linked to libSDL2

FreeSlave:
The client.so must be explicitly linked to libSDL2, just like valve/cl_dlls/client.so.
Related issue: https://github.com/YaLTeR/OpenAG/issues/181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,6 +203,7 @@ elseif(UNIX)
 		# I wasn't able to do this in any other way.
 		# Things like imported targets result in linking with relative path.
 		set(VGUI_LINK_FLAGS "-L${CMAKE_SOURCE_DIR}/linux/release -l:vgui.so")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -202,7 +202,7 @@ elseif(UNIX)
 	else(APPLE)
 		# I wasn't able to do this in any other way.
 		# Things like imported targets result in linking with relative path.
-		set(VGUI_LINK_FLAGS "-L${CMAKE_SOURCE_DIR}/linux/release -l:vgui.so")
```
YaLTeR, line 205: This is `VGUI_LINK_FLAGS`, do you mind moving SDL flags to their own, properly named variable? Also, should it come after or before `-Wl,--no-as-needed`?

FreeSlave: I set another variable.
As for `-Wl,--no-as-needed` - I'm not sure why it's set in the first place. Probably doesn't matter.

YaLTeR:
Thanks!

PR #185, opened 2023-12-23
Make a new list for hardcoded map series in Discord RPC module and fixed declaration of struct in com_model.h

### LINKED ISSUE #162: RPC: Find substring in list for map_name and apply it for thumbnail
Example: `agtricks_telehop` (**map_name**) -> `agtricks` (**list**)

Reposting to the issue to not lost it into the deep of #ag-dev channel

SmileyAG:
Closes #162

### DIFF
--- a/cl_dll/discord_integration.cpp
+++ b/cl_dll/discord_integration.cpp
@@ -1,9 +1,11 @@
+#include <algorithm>
 #include <chrono>
 #include <cstdint>
 #include <cstring>
 #include <memory>
 #include <string>
 #include <unordered_set>
+#include <cctype>
 
 #include <discord_rpc.h>
 
@@ -24,8 +26,26 @@ namespace discord_integration
 		// This seems to be consistent across PCs.
 		constexpr const char STEAM_APP_ID[] = "17215498729465839686";
 
+		struct case_insensitive_hash
+		{
+			size_t operator()(const std::string& str) const
+			{
+				std::string std_lower = str;
+				std::transform(std_lower.begin(), std_lower.end(), std_lower.begin(), [](unsigned char s) { return std::tolower(s); });
+				return std::hash<std::string>()(std_lower);
+			}
+		};
+
+		struct case_insensitive_equal
+		{
+			size_t operator()(const std::string& str1, const std::string& str2) const
+			{
+				return std::equal(str1.begin(), str1.end(), str2.begin(), [](unsigned char s1, unsigned char s2) { return std::tolower(s1) == std::tolower(s2); });
+			}
+		};
+
 		// Maps for which we have thumbnails.
-		const std::unordered_set<std::string> maps_with_thumbnails {
+		const std::unordered_set<std::string, case_insensitive_hash, case_insensitive_equal> maps_with_thumbnails {
 			"2bfree"s,
 			"8b1_hellinashop"s,
 			"ag_bhop_dungeon"s,
@@ -343,6 +363,11 @@ namespace discord_integration
 					get_map_name(map_name, ARRAYSIZE(map_name));
 					if (map_name[0])
 					{
+						// We specifically don't want to convert the map name that will be shown when hovering over the map icon (presence.largeImageText), so that why it got moved above.
+						presence.largeImageText = map_name;
+
+						gHUD.ConvertToLowerCase(map_name);
+
 						if (maps_with_thumbnails.find(map_name) != maps_with_thumbnails.cend())
 						{
 							presence.largeImageKey = map_name;
@@ -369,8 +394,6 @@ namespace discord_integration
 								}
 							}
 						}
-
-						presence.largeImageText = map_name;
 					}
 
 					// Get the server address.
--- a/cl_dll/discord_integration.cpp
+++ b/cl_dll/discord_integration.cpp
@@ -25,7 +25,7 @@ namespace discord_integration
 		// This seems to be consistent across PCs.
 		constexpr const char STEAM_APP_ID[] = "17215498729465839686";
 
-		// This list was specially created for a map series, let’s say that “hl1_bhop_am” series have a several versions of maps (example: `hl1_bhop_am_beta1`), but in fact those maps are almost identical
+		// This list was specially created for a map series, let's say that "hl1_bhop_am" series have a several versions of maps (example: hl1_bhop_am_beta1), but in fact those maps are almost identical
 		// If the beginning of map name matches with what presented in the list, then we set the same name for the thumbnail as in the list (e.g. map name: hl1_bhop_uc1_beta1, thumbnail name: hl1_bhop_uc1)
 		const std::unordered_set<std::string> map_series_with_thumbnails {
 			"hl1_bhop_am"s,
--- a/cl_dll/discord_integration.cpp
+++ b/cl_dll/discord_integration.cpp
@@ -46,7 +46,8 @@ namespace discord_integration
 			"hl1_bhop_uc2"s,
 		};
 
-		const std::unordered_set<std::string> dm_maps {
+		const std::unordered_set<std::string> maps_with_thumbnails {
+			// DM maps
 			"boot_camp"s,
 			"bootbox"s,
 			"bounce"s,
@@ -70,9 +71,7 @@ namespace discord_integration
 			"subtransit"s,
 			"urethane"s,
 			"vengeance"s,
-		};
-
-		const std::unordered_set<std::string> bhop_maps {
+			// Bhop maps
 			"2bfree"s,
 			"8b1_hellinashop"s,
 			"ag_bhop_dungeon"s,
@@ -359,7 +358,7 @@ namespace discord_integration
 
 						convert_to_lower_case(map_name);
 
-						if ((dm_maps.find(map_name) != dm_maps.cend()) || (bhop_maps.find(map_name) != bhop_maps.cend()))
+						if (maps_with_thumbnails.find(map_name) != maps_with_thumbnails.cend())
 						{
 							presence.largeImageKey = map_name;
 						}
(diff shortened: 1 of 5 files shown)


YaLTeR:
> Disabled by default, but can be enabled using discord_rpc_closest_map_match 1

What's the point? Nobody will know about this or use it. Is there any problem that prevents it from being always on? Any map name collisions?

cl_dll/discord_integration.cpp
```diff
@@ -24,8 +26,26 @@ namespace discord_integration
 		// This seems to be consistent across PCs.
 		constexpr const char STEAM_APP_ID[] = "17215498729465839686";
 
+		struct case_insensitive_hash
```
YaLTeR, line 29: Why add this much more complex solution instead of what was already here? Just store the original map_name in a variable before converting to lowercase for the rest of the code.

SmileyAG: Isn't that clear? Imagine that in the list, for example, a contributor writes the name of the map, say, `aGtRiCkS`, when in fact you have an `agtricks` map loaded in the game, and what will you do next? Should we manually change the list every time if that kind of thing happen? That seems like too much to do it each time, so that's that the reason why is such complex solution here

YaLTeR: `// Maps in this list must be lowercase`, then I look at it during review and ask to change

SmileyAG: Uhhhhh... Well, at least in your defense, can be said that selecting entire list and converting it to lower case in IDE shouldn't take than a few seconds to do, so it’s might be really not worth to going for such complex solution

SmileyAG:
> > Disabled by default, but can be enabled using discord_rpc_closest_map_match 1
> 
> What's the point? Nobody will know about this or use it. Is there any problem that prevents it from being always on? Any map name collisions?

Let me give you a couple of examples of why this should be optional:

# Example 1

Let's say we have a series of maps starting with `hl1_bhop_uc1`, say: `hl1_bhop_uc1`, `hl1_bhop_uc1_beta1`, `hl1_bhop_uc1_beta2` and etc

For the most, these are the same maps, but with each version there a few differences, say in the placement / size of triggers or change of map geometry in a few places

So, imagine that in our list there will only be a thumbnail uploaded and filled only for `hl1_bhop_uc1` (final version of map), but we will say that we are playing on a specific version of this series of map (`hl1_bhop_uc1_beta5`), therefore we will not have a thumbnail for it with the current code

This is where `discord_rpc_closest_map_match 1` comes to the rescue
And don’t need to tell me about the hardcode of the list of these series of maps, doing this would require to update client each time and also it took time for fill them

# Example 2

Let's say that there is would be two maps: `kz_synergy` and `kz_synergy_x`, same in naming but completely different in map layout (one is more like bhop-stylish, other is kz-stylish)

Imagine that there is would be thumbnail only for `kz_synergy`, therefore it would also set a thumbnail for `kz_synergy_x` if `discord_rpc_closest_map_match 1` is enabled, while in fact that are completely different map so it would be wrong to set it

YaLTeR:
> This is where discord_rpc_closest_map_match 1 comes to the rescue

Almost certainly nobody will bother to flip a cvar every time they happen to load a map with a problem like this (let alone know which maps are problematic).

> Imagine that there is would be thumbnail only for kz_synergy, therefore it would also set a thumbnail for kz_synergy_x if discord_rpc_closest_map_match 1 is enabled, while in fact that are completely different map so it would be wrong to set it

Well, yeah. Is this going to be a common problem? If so then it might be better to not do this loose matching at all.

SmileyAG:
Just you know, adding an thumbnail for each version of the same map seems like a not smart decision in general, but there is also a issue with it, that in fact you can load a maximum of 300 images per API

Here I see only three options for how to generally solve this issue:

- Hardcode that specific map series (e.g. if map starts with `hl1_bhop_uc1`, then we would always set a thumbnail from `hl1_bhop_uc1` map)
- Make support to configurate from the user’s side, example in some .txt you can type the names of the map series (hl1_bhop_uc1, hl1_bhop_oar and etc) and on the code side you read this file and do the same thing as described in the first option
- `discord_rpc_closest_map_match 1` (but of course it can give the wrong result depends of the specific case as I described above, but this is generally the expected result)

But in general, it’s not to say that there are just a lot of such map series, so maybe the idea of hardcoding this makes sense here

Just make a new list for which the code will also automatically perform a closest match search as it with `discord_rpc_closest_map_match 1` enabled

Example:

```cpp
// This list was specially created for a map series, let’s say that “hl1_bhop_am” series have a several versions of maps (example: `hl1_bhop_am_beta1`), but in fact those maps are almost identical
// If the beginning of map name matches with what presented in the list, then we set the same name for the thumbnail as in the list (e.g. map name: hl1_bhop_uc1_beta1, thumbnail name: hl1_bhop_uc1)
const std::unordered_set<std::string> map_series_with_thumbnails {
	"hl1_bhop_am"s,
	"hl1_bhop_bp1"s,
	"hl1_bhop_bp2"s,
	"hl1_bhop_faf"s,
	"hl1_bhop_lc"s,
	"hl1_bhop_oar"s,
	"hl1_bhop_ocwgh"s,
	"hl1_bhop_pu"s,
	"hl1_bhop_qe"s,
	"hl1_bhop_rp"s,
	"hl1_bhop_st"s,
	"hl1_bhop_uc1"s,
	"hl1_bhop_uc2"s,
};
```

YaLTeR:
Yeah, hardcoding like this sounds like a reasonable approach in this case. And it can be always enabled without a cvar.

cl_dll/discord_integration.cpp
```diff
@@ -24,7 +25,25 @@ namespace discord_integration
 		// This seems to be consistent across PCs.
 		constexpr const char STEAM_APP_ID[] = "17215498729465839686";
 
-		// Maps for which we have thumbnails.
+		// This list was specially created for a map series, let�s say that �hl1_bhop_am� series have a several versions of maps (example: `hl1_bhop_am_beta1`), but in fact those maps are almost identical
```
YaLTeR, line 28: What happened to the quotes here?

cl_dll/hud.cpp
```diff
@@ -937,3 +937,12 @@ float CHud::GetSensitivity( void )
 	return m_flMouseSensitivity;
 }
 
+void CHud::ConvertToLowerCase(const char *str)
```
YaLTeR, line 940: This doesn't need to be a class method (let alone a non-static one). I think one place where I put auxiliary functions was https://github.com/YaLTeR/OpenAG/blob/master/cl_dll/cl_util.h, could be moved there.

cl_dll/discord_integration.cpp
```diff
@@ -343,17 +349,26 @@ namespace discord_integration
...
 						if (maps_with_thumbnails.find(map_name) != maps_with_thumbnails.cend())
+						{
 							presence.largeImageKey = map_name;
-
-						presence.largeImageText = map_name;
+						}
+						else
+						{
+							auto closest_match = std::find_if(map_series_with_thumbnails.begin(), map_series_with_thumbnails.end(), [&](const std::string& str) {
+    												return static_cast<std::string>(map_name).find(str) != std::string::npos; // hl1_bhop_am_beta6 -> hl1_bhop_am
```
YaLTeR, line 364: This finds not just the prefix, but any substring, right? Is that intended?

SmileyAG: > This finds not just the prefix, but any substring, right?

That's not true? Here's an example that I made: https://godbolt.org/z/7Yh6xsWqv

But if you don't like this option, you can of course use one of the next followings to implement that too:

- std::mismatch
- std::equal
- .compare

YaLTeR: It finds with prefix too:

image

But we only want with suffix, right?

SmileyAG: Damn, how I did not thinked about such case at first, yeah then you are right, I would to fix it with one of options that I mentioned above then

cl_dll/discord_integration.cpp
```diff
@@ -24,8 +25,54 @@ namespace discord_integration
...
+			"hl1_bhop_ocwgh"s,
+			"hl1_bhop_pu"s,
+			"hl1_bhop_qe"s,
+			"hl1_bhop_rp"s,
+			"hl1_bhop_st"s,
+			"hl1_bhop_uc1"s,
+			"hl1_bhop_uc2"s,
+		};
+
+		const std::unordered_set<std::string> dm_maps {
```
YaLTeR, line 49: It's better to keep these all in one map, with a comment like `// DM maps` inside. To save one map lookup.

cl_dll/discord_integration.cpp
```diff
@@ -24,8 +25,54 @@ namespace discord_integration
...
 		constexpr const char STEAM_APP_ID[] = "17215498729465839686";
 
-		// Maps for which we have thumbnails.
-		const std::unordered_set<std::string> maps_with_thumbnails {
+		// Maps in next lists must be lowercase!
+		// Discord RPC allow to upload 300 assets per app, but we also use one thumbnail as default icon, so the lists cannot contain more than 299 maps in total!
+
+		// This list was specially created for a map series, let's say that "hl1_bhop_am" series have a several versions of maps (example: hl1_bhop_am_beta1), but in fact those maps are almost identical
+		// If the beginning of map name matches with what presented in the list, then we set the same name for the thumbnail as in the list (e.g. map name: hl1_bhop_uc1_beta1, thumbnail name: hl1_bhop_uc1)
+		const std::unordered_set<std::string> map_series {
```
YaLTeR, line 33: We do a linear search over this so it's better to make it a `vector`.

YaLTeR:
Thanks

PR #186, opened 2024-01-02
Replace cuboid with polygons drawing for triggers on hardware engine if the model data available

SmileyAG:
- Inherit `msurface_t` for a new struct in both ways (depends if the file is C or C++)
- Find engine build number from `sv_version` cvar that is declared in engine-side (that code is taken from **BugfixedHL-Rebased**)
- To the message above, `sv_version` outputs information like this: `sv_version "1.1.2.2/Stdio,48,9920"`
- Replace cuboid with polygons drawing for triggers on hardware engine if the model data available, otherwise continue to draw as cuboid

Yes, it turns out there are enough real examples when triggers may not be a cuboid, so drawing with polygons is seem always a matter of time
But as we already know, in map compilers that are not from Valve, as an optimization it's cut off all of that data for unused brush models and therefore that data is not available

Of course that some tools like bspguy can restore that information from a .bsp file, but I don’t I think that we want now to make a large complex work for such minor task 

This code is not completely final, because I’m still thinking about what exactly speed / offset I want to use for change the alpha in these polygons, after this it most likely will be ready for merge, so for now I just made a pull request for a review but not as a merge

### DIFF
--- a/cl_dll/tri.cpp
+++ b/cl_dll/tri.cpp
@@ -42,6 +42,8 @@ extern IParticleMan *g_pParticleMan;
 #undef min
 #undef max
 
+extern engine_studio_api_t IEngineStudio;
+
 /*
 =================
 HUD_DrawNormalTriangles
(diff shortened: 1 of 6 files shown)


cl_dll/tri.cpp
```diff
@@ -105,9 +107,47 @@ void DrawAACuboid(triangleapi_s *pTriAPI, Vector corner1, Vector corner2)
...
+	if (model && !model->nummodelsurfaces)
+	{
+		DrawAACuboid(pTriAPI, corner1, corner2);
+		return;
+	}
+
+	#ifndef SOFTWARE_BUILD
+	if (IEngineStudio.IsHardware() && model && model->nummodelsurfaces)
+	{
+		#define DrawPolygons(surfs) \
```
tmp64, line 123: That's cursed. Put it in a lambda with `auto` parameter (C++14, which OpenAG builds with) or a template function

YaLTeR:
For some reason from my BXT trigger drawing experiments I remember that the compilers strip out the model data completely, leaving just the collision. Is that not the case?

PR #190, opened 2024-10-25
Half-Life 25th Anniversary Update

sabianroberts:
This PR merges some changes from the Half-Life 25th Anniversary Update. Changes include [but not limited to]:
- Support for larger scale HUD on HD displays
- Changes to view origin/angles
- "Ghost shots" bug fix

Furthermore, a batch file was added so Windows developers and create the build files with ease.

### DIFF
--- a/.GENERATE WINDOWS BUILD FILES.BAT
+++ b/.GENERATE WINDOWS BUILD FILES.BAT
@@ -1,0 +1,2 @@
+cmake -A Win32 -B build
+pause
--- a/cl_dll/ammo.cpp
+++ b/cl_dll/ammo.cpp
@@ -74,10 +74,14 @@ void WeaponsResource :: LoadWeaponSprites( WEAPON *pWeapon )
 {
 	int i, iRes;
 
-	if (ScreenWidth < 640)
-		iRes = 320;
-	else
+	if (ScreenWidth > 2560 && ScreenHeight > 1600)
+		iRes = 2560;
+	else if (ScreenWidth >= 1280 && ScreenHeight > 720)
+		iRes = 1280;
+	else if (ScreenWidth >= 640)
 		iRes = 640;
+	else
+		iRes = 320;
 
 	char sz[256];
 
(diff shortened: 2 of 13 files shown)


.GENERATE WINDOWS BUILD FILES.BAT
```diff
@@ -0,0 +1,2 @@
+cmake -A Win32 -B build
```
YaLTeR, line 1: I don't think this file is needed

YaLTeR (commented):
Do we want to keep a pre-25th anniversary update build? In that case the changes should be under a new define.

sabianroberts:
> Do we want to keep a pre-25th anniversary update build? In that case the changes should be under a new define.

Since there's no serverside changes here, I don't think a pre-25th anniversary build is needed. The most notable change here is the larger HUD which is only going to benefit players especially those on 1440p+ screens. Furthermore 99% of mods are fully compatible with the HL25 engine. 

Unless AG's full mod package has its own "hud.txt" the one from the valve folder is used and the HUD scaling works correctly.

tmp64:
> Unless AG's full mod package has its own "hud.txt" the one from the valve folder is used and the HUD scaling works correctly.

It does and that file needs to be updated

image

sabianroberts:
Thanks for noting! My ag/sprites/hud.txt got deleted somehow 🤔

larziwau:
> Do we want to keep a pre-25th anniversary update build? In that case the changes should be under a new define.

well we can create a branch for it?

YaLTeR:
I think if two builds are needed then a define would be better to avoid having to update both versions of the code every time.

sabianroberts:
Somewhere in the code you could somehow get the EXE version of the game and use strcmp to make an if statement where if version >= HL25, define `HL25` and use `#ifdef HL25` where applicable.

Glitch31415:
just found AG, does it still not support the HL25 client?

sabianroberts:
@Glitch31415 the game works fully on the HL25 client but its codebase hasn't been updated to match Half-Life 1's new features such as the larger scale HUD.

PR #205, opened 2025-07-12
Remove XP build from Continuous Integration workflow

sabianroberts:
GitHub Actions has completely deprecated the Windows-2019 image, as well as the v141_xp compilers having been deprecated since VS2017. This has been causing OpenAG's workflows to be considered "failed" for a while now. 

While I doubt anyone needs a Windows XP-compatible version of OpenAG anymore, if one were to really need one, it can easily be built locally.

This PR removes the v141_xp job and sets the v143 job to `windows-latest`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -115,3 +115,10 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y clang-${{ matrix.clang-ver }} g++-multilib libgl-dev ninja-build rapidjson-dev
+          
+      - name: CMake generate
+        run: cmake -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.configuration }} -B build
+
+      - name: Build
+        working-directory: build
+        run: ninja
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -35,8 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - { runs-on: windows-2019, toolset: v141_xp }
-          - { runs-on: windows-2022, toolset: v143 }
+          - { runs-on: windows-latest, toolset: v143 }
```
YaLTeR, line 38:
```suggestion
          - { runs-on: windows-2022, toolset: v143 }
```
I prefer bumping these manually when they go deprecated so that the environment doesn't change under me

.github/workflows/CI.yml
```diff
@@ -117,10 +115,3 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y clang-${{ matrix.clang-ver }} g++-multilib libgl-dev ninja-build rapidjson-dev
-
```
YaLTeR, line 120: Why was this removed?

sabianroberts: my apologies, I removed this by mistake.

.github/workflows/CI.yml
```diff
@@ -117,7 +115,7 @@ jobs:
...
+
```
YaLTeR, line 118:
```suggestion

```

YaLTeR:
Thanks

